PR #25, opened 2020-02-04
Mobile Nav

spbyrne:
V1 Mobile nav.

Required some reshuffling of docs specific components. The docs pages are now using their own layout & header to retain more control without affecting the rest of the website. It also allowed me to remove props from the layout component to simplify things.

On marketing pages you get the four page links but search is hidden until I write mobile styles (previously it was hidden as well). On the docs pages you get the docs nav but no page nav to go back to the marketing pages aside from tapping Tina to go to the home page. I'll add the regular links above the docs nav but for now this works.

The header is fixed on the docs pages. I'm on the fence about this. Maybe if we hide the nav/github icon when scrolling and display them when you scroll up it would look a bit more tidy.

### DIFF
--- a/pages/blog/[slug].tsx
+++ b/pages/blog/[slug].tsx
@@ -31,6 +31,7 @@ export default function BlogTemplate(props) {
 
 BlogTemplate.getInitialProps = async function(ctx) {
   const { slug } = ctx.query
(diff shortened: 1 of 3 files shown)


DirtyF:
Do you plan to revisit mobile layout to make search available for resolutions < 620 px?

PR #50, opened 2020-02-10
Stylefix

spbyrne:
* Creates blog pagination component, adds styling
* Improve font fallback behaviour
* Add react memo to docs layout & header nav
* Adjust various styles
* Add iframe styles to richtext
* Closes docs section on nav (if no longer in that section)
* Reduce homepage code size on mobile


components/ui/BlogPagination.tsx
```diff
@@ -0,0 +1,127 @@
...
+    const prevPage =
+      currentPage - 1 === 1 ? '/blog' : `/blog/page/${currentPage - 1}`
+    const nextPage = `/blog/page/${currentPage + 1}`
+    const [selectValue, setSelectValue] = useState(currentPage)
+
+    function handleSelectChange(e) {
+      e.preventDefault()
+      const pageNumber = e.target.value
+      setSelectValue(pageNumber)
+      if (pageNumber === '1') {
```
jamespohalloran: I don't think we need this conditional, as /blog/page/1 should work

components/ui/BlogPagination.tsx
```diff
@@ -0,0 +1,127 @@
...
+import Router from 'next/router'
+
+import { DynamicLink } from './DynamicLink'
+
+export const BlogPagination = styled(
+  ({ currentPage, numPages, ...styleProps }) => {
+    const isFirst = currentPage === 1
+    const isLast = currentPage === numPages
+    const prevPage =
+      currentPage - 1 === 1 ? '/blog' : `/blog/page/${currentPage - 1}`
```
jamespohalloran:
```suggestion
      `/blog/page/${currentPage - 1}`
```
/blog/page/1 should still work

components/ui/BlogPagination.tsx
spbyrne: Tested and indeed that works, but I do like that this keeps the URL clean (/blog vs. /blog/page/1). I'm fine with removing it, though. Simplifies things.

spbyrne: @kendallstrautman What do you think?

kendallstrautman: yeah that's fine as long as `/blog` still works

kendallstrautman: ill fix this in another branch

PR #56, opened 2020-02-11
Adds inline editing

kendallstrautman:
This PR **adds rudimentary Inline Editing to all pages**. I moved over a few inlined editing fields from the monorepo in order to have more control over the styling. I also added a few new inline fields: `InlineWysiwyg`, `InlineTextArea`, & `BlockTextArea` — those can be found in `components/layout/ui/inline`.

### Other Notes: 

There's also some refactoring / cleanup, but nothing major — mostly just removing unused imports and such. **We need to stay on `"next": "9.2.2-canary.12"`** until the newer releases implement the ['fallback' behavior](https://github.com/zeit/next.js/issues/9524#issuecomment-584710327) described in the RFC.

### Next steps for Inline Editing implementation: 
 - _Make it pretty_ 💅, this goes beyond just adding styles but really digging into the UX/DX to see how to best interact with this inline editing API. Talks about longer term 'global toolbar' stuff will probably need to come back on the table. 
- We have _quite a few inline fields and inline block fields to make_. The first one off the top of my head would be group-list, just cause I really needed that on the home and teams pages. 
- Would be cool if we felt inspired to _convert the whole home page into a 'page builder'_ with blocks type situation.

## Update 2/19

All inline editing controls are 'hidden' (i.e. commented out) so that inline editing mode can't be toggled until we are ready to release it to the wild. All the tina forms are setup to work in the sidebar, so we can use that for editing via open authoring.


kendallstrautman:
Tight none of the pages are rendering. Looks like that ` if (!form) return null` is running on the prod build

kendallstrautman:
Still need to: 
- [ ] - Make sure all the sidebar forms are 👌🏼
- [ ] - Ensure the inline editing stuff doesn't render in prod

PR #59, opened 2020-02-11
Media Store Documentation #733

ncphillips:
This PR will provide documentation for the new media APIs introduced in tinacms/tinacms#733


kendallstrautman:
Should we close this since it seems this was [documented](https://tinacms.org/docs/media)?

PR #60, opened 2020-02-11
Docs for Number Field Plugin #726

ncphillips:
Docs for tinacms/tinacms#726


content/docs/fields/number.md
```diff
@@ -0,0 +1,53 @@
...
+interface TextConfig {
+  name: string
+  component: 'number'
+  label?: string
+  description?: string
+  step?: string
+}
+```
+
+-->
```
PaulBunker: Is this an intentional arrow? Also looks like next and previous is broken.

DirtyF:
Nice, here's an image for the number field if it hellps 😊 

number-field

content/docs/fields/number.md
```diff
@@ -0,0 +1,53 @@
...
+consumes:
+  - file: /packages/tinacms/src/plugins/fields/NumberFieldPlugin.tsx
+    details: Shows text field interface and config options
+  - file: /packages/@tinacms/fields/src/NumberField.ts
+    details: Shows text field interface and config options
+---
+
+The `number` field represents a number input.
+
+<!-- TODO: Provide image of number field -->
```
DirtyF: submitted one in comment :)

content/docs/fields/number.md
ncphillips: Nope! good catch

ncphillips:
Okay, I think I fixed up those issues. Thanks @DirtyF and @PaulBunker

content/docs/fields/number.md
PaulBunker: It actually looked quite pretty on the preview site 😂

PR #92, opened 2020-02-13
Revise cms docs

dwalkr:
Includes new/updated, react-based, SSG-agnostic docs for:

- [x] CMS object / initialization
- [x] Forms
- [x] Inline Editing
- [x] Media

other todos
- [x] Determine appropriate redirects for removed docs
- [x] Add consumer metadata to new docs


content/docs/fields.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Field Config
+
+All field plugins share a common config:
+
+```typescript
+interface FieldConfig {
+  name: string
+  parse(): any
+  format(): any
```
ncphillips: Missing the parameters

content/docs/cms.md
```diff
@@ -0,0 +1,296 @@
...
+> }
+> ```
+
+## CMS Configuration
+
+When instantiating the `TinaCMS` object, you can pass in a configuration object. This allows you to configure some options for the sidebar, and also allows you to configure Plugins and APIs declaratively.
+
+```typescript
+interface TinaCMSConfig {
+  plugins?: Array<Plugin>
```
ncphillips:
```suggestion
  plugins?: Plugin[]
```

content/docs/forms.md
```diff
@@ -0,0 +1,214 @@
...
+interface FormOptions<S> {
+  id: any
+  label: string
+  fields: Field[]
+  loadInitialValues?: () => Promise<S>
+  reset?(): void
+  actions?: any[]
+  meta?: {
+    [key: string]: string
+  }
```
ncphillips: I don't think this is worth mentioning.

```suggestion
```

ncphillips: In fact, I don't know why this exists. This should be removed from `Form`

content/docs/forms.md
```diff
@@ -0,0 +1,214 @@
...
+
+> Note that when using these basic hooks, you are expected to implement the save functionality yourself by adding an `onSubmit` function. By default, Tina makes no assumptions about how your content is stored. These basic form hooks are building blocks for creating more purpose-built tools to fit specific use cases.
+
+### Watched Vars: Preserve Form Reactivity
+
+The second argument that can be passed to the form hooks is an optional object containing values that the form will need to react to. Use this when dealing with data from external sources or in an environment with Hot Module Replacement; in other words, when you expect this data to change and it's essential to keep the form in sync with it.
+
+```typescript
+interface WatchableFormValue {
+  values: any
```
ncphillips: Maybe we should say:
```suggestion
  values: object
```

content/docs/forms.md
```diff
@@ -0,0 +1,214 @@
...
+  label: FormOptions<any>['label']
+  fields: FormOptions<any>['fields']
```
ncphillips: Duplication in the docs is okay. Makes things more readable.

```suggestion
  label: string
  fields: Field[]
```

dwalkr: for sure, this is better. I just copypasted these interfaces from the source

content/docs/forms.md
dwalkr: I think this was originally how we were gonna do global forms at one point. I'll remove any reference to meta

PR #93, opened 2020-02-13
Add field plugin post

kendallstrautman:
Still todo: 
- [x] update the demo repo to reflect the new `_app.js` changes
- [x] test the code examples

### DIFF
--- a/content/blog/custom-field-plugins.md
+++ b/content/blog/custom-field-plugins.md
@@ -1,6 +1,6 @@
 ---
 title: Creating Field Plugins
-date: '2020-02-14T07:00:00.000Z'
+date: '2020-02-17T07:00:00.000Z'
 author: Kendall Strautman
 draft: false
 consumes:
--- a/content/blog/custom-field-plugins.md
+++ b/content/blog/custom-field-plugins.md
@@ -16,17 +16,17 @@ In the [previous post](https://tinacms.org/blog/custom-field-components), we lea
 
 ## Field Plugin vs. Field Component
 
-Plugins **extend functionality in the CMS**; field plugins allow us to create and register custom fields. _There are a few reasons why you may want to create a field plugin._ With a custom field, you can completely control the editing experience and functionality. If the basic fields provided by Tina don't fit your use case, **you can fill the gaps**.
+Plugins **extend functionality in the CMS**; field plugins allow us to create and register custom fields. _There are a few reasons why you may want to create a field plugin._ With a custom field, you can completely control the editing experience and functionality. If the primary fields provided by Tina don't fit your use case, **you can fill the gaps**.
 
 A field component is just one piece of a field plugin ([more on this later](https://tinacms.org/blog/custom-field-plugins#field-plugin-interface)). A custom field component can achieve the same functionality as a plugin. But if you plan on reusing the custom field on different forms, it is recommended to take the extra steps to make a plugin 🔌.
 
-Creating field plugins helps confine complex logic to a single module. This makes it easier to update or swap out custom field functionality, later down the line. Using **the plugin API makes our 'higher-level' code more reusable** and contained, keeping the fields independent from the core CMS.
+Creating field plugins helps confine complex logic to a single module. This makes it easier to update or swap out custom field functionality later down the line. Using **the plugin API makes our 'higher-level' code more reusable** and contained, keeping the fields independent from the core CMS.
 
 > For those who want to dig deeper, this approach seeks to embody the [Dependency Inversion Principle](https://stackify.com/dependency-inversion-principle/).
 
 ## Getting Started 👏
 
-To follow along, you should have a custom field component set-up with a Tina form. If not, you can get some more context from the previous post: [how to create a custom field component](https://tinacms.org/blog/custom-field-components). In the following examples, I will continue using the [llama-filters](https://github.com/kendallstrautman/llama-filters) 🦙 demo from that post.
+To follow along, you should have a custom field component set-up with a Tina form. If not, you can get some more context from the previous post: [how to create a custom field component](https://tinacms.org/blog/custom-field-components). In the following examples, I am using the same [llama-filters](https://github.com/kendallstrautman/llama-filters) 🦙 demo from our previous post.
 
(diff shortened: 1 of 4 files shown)


content/blog/custom-field-plugins.md
```diff
@@ -0,0 +1,144 @@
...
+consumes:
+  - file: /packages/next-tinacms-json/src/use-local-json-form.ts
+    details: Example shows useLocalJsonForm in use
+  - file: /packages/next-tinacms-json/src/use-json-form.ts
+    details: Example shows useLocalJsonForm in use
+  - file: /packages/@tinacms/form-builder/src/field-plugin.tsx
+    details: Depends on the FieldPlugin interface
+---
+
+In the [previous post](https://tinacms.org/blog/custom-field-components), we learned how to create a custom field component and register it to the sidebar. So let's **go full circle on the topic of custom fields** in TinaCMS. In this short but sweet post we’ll cover how to _turn a plain ol’ field component into a field plugin._
```
DirtyF: Why would we want to do that?

content/blog/custom-field-plugins.md
```diff
@@ -0,0 +1,144 @@
...
+}
+```
+
+You’ll want to import the custom field component and then register the plugin with the CMS directly. Notice how we import the `RangeInput` component created in the [previous post](https://tinacms.org/blog/custom-field-components), this is the custom _field component_ that we're now attaching to a _field plugin_.
+
+> If you’re working with Gatsby, this [looks slightly different](https://tinacms.org/docs/gatsby/custom-fields/). _Hint_: you’ll head to the **gatsby-browser.js** file to access the CMS instance.
+
+### Field Plugin Definition
+
+A plugin is used to extend functionality in the CMS. A field plugin in particular allows us to create and register custom fields to the CMS.
```
DirtyF: OK, but why?

content/blog/custom-field-plugins.md
```diff
@@ -0,0 +1,144 @@
...
+    meta: any,
+    field: Field
+  ): string | object | undefined
+  parse?: (value: any, name: string, field: Field) => any
+  format?: (value: any, name: string, field: Field) => any
+  defaultValue?: any
+}
+```
+
+You can see that there are additional configuration functions and options. These options are incredibly useful for creating more complex fields that require [validation](https://tinacms.org/docs/fields/custom-fields#validate-optional) or formatting, such as an [email field](https://tinacms.org/docs/gatsby/custom-fields/).
```
DirtyF: ... finally here is the explanation 😄We should sum up this in the intro to give a sense of purpose to the article.

content/blog/custom-field-plugins.md
```diff
@@ -0,0 +1,144 @@
...
+    allValues: any,
+    meta: any,
+    field: Field
+  ): string | object | undefined
+  parse?: (value: any, name: string, field: Field) => any
+  format?: (value: any, name: string, field: Field) => any
+  defaultValue?: any
+}
+```
+
```
DirtyF: It's not clear to me where to put that definition 😊

content/blog/custom-field-plugins.md
```diff
@@ -0,0 +1,144 @@
...
+You can see that there are additional configuration functions and options. These options are incredibly useful for creating more complex fields that require [validation](https://tinacms.org/docs/fields/custom-fields#validate-optional) or formatting, such as an [email field](https://tinacms.org/docs/gatsby/custom-fields/).
+
+## 2. Use the custom field in a form
+
+Now that the plugin is registered with the CMS, we can use it in any form definition. Going back to the [llama-filters demo](https://github.com/kendallstrautman/llama-filters), let’s head to index.js where the Tina form is configured. We need to **update the form options** for our image saturation field to reference the field plugin as opposed to calling the component directly.
```
DirtyF:
```suggestion
Now that the plugin is registered with the CMS, we can use it in any form definition. Going back to the [llama-filters demo](https://github.com/kendallstrautman/llama-filters), let’s head to `index.js` where the Tina form is configured. We need to **update the form options** for our image saturation field to reference the field plugin as opposed to calling the component directly.
```

content/blog/custom-field-plugins.md
```diff
@@ -0,0 +1,144 @@
...
+      },
+    })
+
+    /*
+     ** 2. Define the field plugin
+     */
+    this.customFieldPlugin = {
+      name: 'range-input',
+      Component: RangeInput,
+    }
```
ncphillips: I might not make the object a property the Site class. There's no real advantage to doing so, and I would bet that people will think it's necessary.

You could just make it a variable scoped to the constructor:

```ts
constructor() {
  // ...
  let rangeFieldPlugin = {
    name: 'range-input',
    Component: RangeInput
  }
  // ...
}
```

or you could make it a module-scoped variable:

```
let rangeFieldPlugin = {
  name: 'range-input',
  Component: RangeInput
}

export default class Site extends App {
  // ...
```

Personally, I prefer the constructor scope for this blog post because it maintains the order for your instructions.

content/blog/custom-field-plugins.md
ncphillips: I would switch from "definition" to "interface". What's described above is the interface for field plugins. You don't have to put it anywhere, it already exists in the tinacms codebase.

content/blog/custom-field-plugins.md
ncphillips: Note that it's still possible to put validate, parse, and format on other fields. The usefulness of FieldPlugins comes from not having to do that every single time. 

For instance, why do we add a "date" field plugin instead of just making the `DateField` component available? Because we want to make sure that the `parse` and `format` functions are always supplied.

```ts
{ 
  name: "created_at", 
  component: "date" 
}
```
is much nicer than:

```ts
import { DateField, parseDate, formatDate} from "tinacms"

{ 
  name: "created_at", 
  component: DateField, 
  parse: parseDate,
  format: formatDate,
},
```

An additional note (that might be too heady for this blog post) is that this approach better adheres to the Dependency Inversion Principle. By using the plugin API to reference the "date" field, the modules that describe the forms do not rely on some concrete implementation of a date field. If you find a new date field, you just update or replace the "date" plugin and it updates all your forms.

kendallstrautman:
Thanks for the feedback! Made some changes but I'd like to read through it again on Monday ☺️

PR #96, opened 2020-02-14
Documentation: Update example code on Gatsby manual setup page.

raubin:
This adds one more config item to the existing code sample on the [Gatsby manual setup](https://tinacms.org/docs/gatsby/manual-setup) page, which appears to be required.

When following the current documentation, adding the `gatsby-plugin-tinacms` plugin to your `gatsby-config.js` results in this error: 

`Unhandled Rejection (TypeError): Cannot read property 'position' of undefined`

Adding plugin options for sidebar position will fix this and meet the description of this first step of manual setup.

I've commented the `hidden` key out since it's not required for things to work, but might make sense to keep since it hints at how you'd probably want this configured for a real project.

Hope this helps!

### DIFF
--- a/content/docs/gatsby/manual-setup.md
+++ b/content/docs/gatsby/manual-setup.md
@@ -44,7 +44,7 @@ module.exports = {
       resolve: 'gatsby-plugin-tinacms',
       options: {
         sidebar: {
-          // hidden: process.env.NODE_ENV === "production",
+          hidden: process.env.NODE_ENV === "production",
           position: "displace",
         },
         plugins: [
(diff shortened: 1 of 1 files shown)


content/docs/gatsby/manual-setup.md
```diff
@@ -43,6 +43,10 @@ module.exports = {
     {
       resolve: 'gatsby-plugin-tinacms',
       options: {
+        sidebar: {
+          // hidden: process.env.NODE_ENV === "production",
```
kendallstrautman:
```suggestion
          hidden: process.env.NODE_ENV === "production",
```

PR #103, opened 2020-02-18
Ssr previews validation and checks

jhuggett:


pages/github/fork.tsx
```diff
@@ -22,15 +24,72 @@ export default function Authorizing() {
     }
   }
 
+  const [forkValidating, setForkValidating] = useState(true)
+ 
+  useEffect(() => {  
+    const branch = "master";
+    fetch(`/api/proxy-github`, {
+      method: 'POST',
```
jamespohalloran: method 'POST', but proxy method 'GET'

jhuggett: I get `TypeError: Failed to execute 'fetch' on 'Window': Request with GET/HEAD method cannot have body.` when I use 'GET' on the proxy.

jamespohalloran:
Cool! That stuff looks good. I might suggest making a few changes to make the fork checking a bit simpler. 
The check for a fork & branch happens both within fork.tsx and EditLink; could w epull that out into it's own helper maybe? 

Also the nested `.then` statements in `fork.tsx` are a bit hard to read. It might be cleaner if you switch to using async/await

PR #115, opened 2020-02-20
Removes editing context, uses `cms.sidebar.hidden` flag

kendallstrautman:
~~Not sure why there's so many merges 😑. But this _should_ work. I couldn't properly test it since I don't have the open authoring stuff setup yet, but toggling the `editMode` values seem to work right.~~ nvm not working rn lol

**Update 2/24** - We had to set `sidebar.hidden` in a timeout 🤷‍♀️


pages/_app.tsx
```diff
@@ -7,60 +7,56 @@ import { DefaultSeo } from 'next-seo'
...
   const tinaConfig = {
     cms: {
       apis: {
         git: new GitClient('http://localhost:3000/___tina'),
       },
     },
     sidebar: {
-      hidden: !isEditMode,
+      // editMode initially set here
+      hidden: process.env.NODE_ENV === 'production' && true,
```
ncphillips: I assume `&& true` is here to make it a boolean That's a fine approach, here's another:

```suggestion
      hidden: !!(process.env.NODE_ENV === 'production'),
```

pages/index.tsx
```diff
@@ -18,15 +18,15 @@ import {
...
 import { InlineTextareaField, BlockTextArea } from '../components/ui/inline'
 import { useLocalGithubJsonForm } from '../utils/github/useLocalGithubJsonForm'
 import getJsonData from '../utils/github/getJsonData'
 import { getGithubDataFromPreviewProps } from '../utils/github/sourceProviderConnection'
 
 const HomePage = (props: any) => {
-  const editContext = useEditContext()
-  editContext.setIsEditMode(props.editMode)
+  // sets cms.sidebar.hidden based on editMode
+  setIsEditMode(props.editMode)
```
ncphillips: I have a vague feeling this might cause an infinite loop, but I'm not sure 🤔

kendallstrautman:
Current state - https://www.loom.com/share/e2bc6088415a4ac1ad50daf0f253449d - it _seems_ like the hidden flag is being set okay when toggling edit modes...but getting some weird behavior where the edit icon still shows up. I tested the pre-release on a separate next project and couldn't reproduce so not sure what is going on. The latest commit has where I left off...

jamespohalloran:
This one is proving to be a doozy.

One workaround is to set hidden in a timeout
```
    setTimeout(() => {
      cms.sidebar.hidden = !props.editMode
    })
```
But I suspect there's an underlying bug setting a subscribable on the first render?
Nolan had this working properly on a stripped-down Next site though.

ncphillips:
Yeah. This works fine in `demo-next`. I'm not sure if it's something about the way site is setup, or if it's something with the new version of `next`

kendallstrautman:
UGHHHH its broken again after merging ssr-previews

kendallstrautman:
Should be good now?

PR #145, opened 2020-02-25
2020 03 02 release docs

ncphillips:
[Milestone](https://github.com/tinacms/tinacms/milestone/14?closed=1)

### DIFF
--- a/content/docs/fields/html.md
+++ b/content/docs/fields/html.md
@@ -1,28 +1,29 @@
 ---
-title: Markdown Field
-prev: /docs/fields/textarea
-next: /docs/fields/html
+title: HTML Field
+prev: /docs/fields/markdown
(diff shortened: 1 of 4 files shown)


content/docs/fields/html.md
```diff
@@ -0,0 +1,49 @@
...
+    },
+    // ...
+  ],
+}
+```
+
+## Options
+
+- `name`: The path to some value in the data being edited.
+- `component`: The name of the React component that should be used to edit this field. Available field component types are [defined here](/docs/concepts/fields#field-types)
```
dwalkr:
```suggestion
```

/docs/concepts/fields doesn't exist anymore and I think including `component` as one of the options  is a bit confusing, because if component is anything other than 'html' then it's a different field

PR #152, opened 2020-02-26
Deleted fork and autoAuth

jhuggett:
ADDED: 
- Going to the homepage with "autoAuth" query param opens a modal to begin authentication right away.
- When the fork is deleted and the homepage is refreshed, a modal pops up saying fork is gone.

TODO: 
- Add this functionality to the other pages
- Handle when content is saved not just when content is gotten


pages/index.tsx
```diff
@@ -110,8 +115,52 @@ const HomePage = (props: any) => {
     props.editMode
   )
 
+  const [autoAuthPopupShow, setAutoAuthPopupShow] = useState(false)
```
jamespohalloran: maybe:
`const [authPopupDisplayed, setAuthPopupDisplayed] = useState(false)`

pages/index.tsx
```diff
@@ -110,8 +115,52 @@ const HomePage = (props: any) => {
...
+
+  const refreshPage = () => {
+    fetch(`/api/reset-preview`).then(() => {
+      window.location.href = "/?autoAuth"
+    })
+  }
+
+  useEffect(() => {
```
jamespohalloran: This functions pretty simple, I'd probably just:
```  
useEffect(() => {
      if (window.location.href.includes("autoAuth")) {
        setAuthPopupDisplayed(true)
      }
  })
```

pages/index.tsx
```diff
@@ -110,7 +115,47 @@ const HomePage = (props: any) => {
...
+    <InlineForm form={form}>
+      {authPopupDisplayed && (
+        <ActionableModal title="Authentication" message="To edit this site, you first need to be authenticated." actions={[
+          {
+            name: "Continue",
+            action: enterEditMode,
+          },
+          {
+            name: "Cancel",
+            action: null,
```
jamespohalloran: It's a bit weird that we do a check for autoAuthPopupShow here, and also in side the modal itself. Maybe instead, the action passed could be 
```
() => {
setAuthPopupDisplayed(false)
}
```
Then the modal is no longer responsible for a separate open/closed state

pages/index.tsx
```diff
@@ -110,7 +115,47 @@ const HomePage = (props: any) => {
...
+  
+  
+
   return (
+    <InlineForm form={form}>
+      {authPopupDisplayed && (
+        <ActionableModal title="Authentication" message="To edit this site, you first need to be authenticated." actions={[
+          {
+            name: "Continue",
+            action: enterEditMode,
```
jamespohalloran: We may want to remove the `autoAuth` query param on load, so that if they accept or cancel, it won't re-pop upon refresh (or on page change)

utils/github/getJsonData.ts
```diff
@@ -9,13 +9,19 @@ const getJsonData = async (
...
     const response = await getContent(
       sourceProviderConnection.forkFullName,
       sourceProviderConnection.headBranch || 'master',
       filePath,
       sourceProviderConnection.accessToken
     )
 
+    if (response.response) {
+      if (response.response.status == 404)
+        return 'Content not found. Your fork may have been deleted.'
```
jamespohalloran: I think the build is failing because this isn't always expected to be a string. Maybe we should throw an error here instead (or return an object with a error prop). We'll have to do the same thing with the getMarkdownData below as well

PR #163, opened 2020-02-28
feat: persist changes to localStorage automatically

dwalkr:
- [x] create localStorage CMS API
- [x] save to localStorage on change
- [x] load from localStorage when entering edit mode
- [x] ~destroy pending changes on save?~ this isn't strictly necessary either; localStorage always stays in sync w/ form state. would re-initialize localStorage state immediately after save so there's no point
- [x] ~customize reset behavior? (might not be necessary; calls `form.updateValues` instead of using `loadInitialValues`)~ deemed unnecessary; resetting `form.values` automatically propagates reset to localStorage

- [x] ~**fix bug:** editing WYSIWYG doesn't trigger `useWatchFormValues` subscription (confirmed: https://github.com/tinacms/tinacms/issues/823 )~ found a workaround
- [x] ~**potential enhancement:** also display pending changes in list view~ too extracurricular if nothing is editable on that screen


pages/blog/[slug].tsx
```diff
@@ -42,6 +43,25 @@ export default function BlogTemplate({
...
+
+  // load from storage on boot
+  useEffect(() => {
+    if (!editMode) return
+
+    const values = cms.api.storage.load(markdownFile.fileRelativePath)
+    if (values) {
+      form.updateValues(values)
+    }
+  }, [editMode])
```
Comment (ncphillips) [inline pages/blog/[slug].tsx]: I suspect that if you navigate between blog the local-storage would fail. You might want to check that. Adding `form` to the deps list would fix that issue
```suggestion
  }, [form, editMode])
```

utils/plugins/BrowserStorageApi.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  debouncePersist() {
+    this.timeout && clearTimeout(this.timeout)
+    this.timeout = setTimeout(this.persist.bind(this), 1000)
+  }
+
+  persist() {
+    this.storage.setItem(this.namespace, JSON.stringify(this.data))
+  }
```
ncphillips: Seems like these want to be private

pages/blog/[slug].tsx
Comment (dwalkr) [reply pages/blog/[slug].tsx]: I think you're right, it just so happens there's no direct navigation between blog posts on the site

components/layout/OpenAuthoringSiteForm.tsx
```diff
@@ -33,6 +35,26 @@ const OpenAuthoringSiteForm = ({
     setTimeout(() => (cms.sidebar.hidden = !editMode), 1)
   }, [])
 
+  /**
+   * persist pending changes to localStorage
+   */
+
+  const saveToStorage = useCallback(formData => {
+    cms.api.storage.save(path, formData.values)
+  }, [])
```
jamespohalloran:
```suggestion
  }, [path])
```

dwalkr: yep, good catch

PR #165, opened 2020-03-02
flags error on content save failure

jhuggett:


jamespohalloran:
Right now useLocalGithubJsonForm,  need to be contextually aware that they are wrapped in a OpenAuthoringModalContainer (or something that handles an openAuthSaveError event). If we replaced the error modal with toast for example, this might break.

I wonder if useLocalGithubJsonForm and the other hooks should bubble the error up, and let a higher up error boundary handle it.

jamespohalloran:
Just pushed up a commit to show an alert onSuccessful save. It uses final-forms decorators: https://github.com/tinacms/tinacms.org/commit/4e8196412898f462c3701fa4713c5c5f7f83ee2f

Here's a good blog post explaining them: https://blog.cloudboost.io/final-form-decorators-calculated-fields-and-warnings-bf9cccd21b7e

There's also an afterSubmitFailed callback as well, I wonder if it makes sense to trigger the modal from here

utils/github/useLocalGithubJsonForm.ts
```diff
@@ -38,9 +39,10 @@ const useGithubJsonForm = <T = any>(
           setCachedFormData(jsonFile.fileRelativePath, {
             sha: response.data.content.sha,
           })
-        })
-        .catch(e => {
-          alert('Something went wrong!')
+        }).catch(e => {
+          if (e.response.status == 404) {
+            return { [FORM_ERROR]: 'A special error message' }
```
jamespohalloran: I think this stub error message text snuck through :) haha

utils/github/useLocalGithubMarkdownForm.ts
```diff
@@ -38,6 +39,10 @@ const useGithubMarkdownForm = <T = any>(
         setCachedFormData(markdownFile.fileRelativePath, {
           sha: response.data.content.sha,
         })
+      }).catch(e => {
+        if (e.response.status == 404) {
+          return { [FORM_ERROR]: 'A special error message' }
```
jamespohalloran: Same here

components/layout/OpenAuthoringSiteForm.tsx
```diff
@@ -43,14 +44,23 @@ const OpenAuthoringSiteForm = ({
     return undecorateSaveListener
   }, [form])
 
+  useEffect(() => {
+    const failedListener = createDecorator({
+      afterSubmitFailed: () => setStatefulPreviewError('Failed to save data. Your fork may have been deleted.')
```
jamespohalloran: This could go in the same useEffect as the one above:
```
    const submitListener = createDecorator({
      afterSubmitSucceeded: () =>
        cms.alerts.success(
          `Saved Successfully: Changes committed to ${Cookies.get(
            'fork_full_name'
          )}`
        ),
afterSubmitFailed: //...
    })
```

utils/plugins/MarkdownCreatorPlugin.ts
```diff
@@ -98,6 +99,10 @@ export class MarkdownCreatorPlugin<FormShape = any, FrontmatterShape = any>
       if (this.afterCreate) {
         this.afterCreate(response)
       }
+    }).catch(e => {
+      if (e.response.status == 404) {
+        return { [FORM_ERROR]: 'A special error message' }
```
jamespohalloran: here too

PR #170, opened 2020-03-02
Blog Creator on base Layout

kendallstrautman:
Now the blog creator plugin is on the base Layout, so it will be accessible on all pages except docs. Had to adjust the redirect url for the newly created file so it worked from any page.


kendallstrautman:
Nolan  1 hour ago
The MarkdownCreator plugin should maybe be inrstantiated in a useMemo.

PR #182, opened 2020-03-04
Fix 'key' warnings

kendallstrautman:
Some of the 'key' warnings need to be solved in Tina. But I fixed the ones I could in the toolbar.

### DIFF
--- a/components/cms/Toolbar.tsx
+++ b/components/cms/Toolbar.tsx
@@ -27,15 +27,25 @@ export const Toolbar = styled(({ ...styleProps }) => {
           <CreateContentMenu />
         </Create>
         <Github>
-          {git.all().length > 0 && git.all().map((git: any) => git.component())}
+          {git.all().length > 0 &&
+            git.all().map((git: any, i) => {
+              const id = `${i}-${git.name}`
+              return git.component(id)
+            })}
         </Github>
         <Status>
           {status.all().length > 0 &&
-            status.all().map((status: any) => status.component())}
+            status.all().map((status: any, i) => {
+              const id = `${i}-${status.name}`
+              return status.component(id)
+            })}
         </Status>
         <Actions>
           {actions.all().length > 0 &&
-            actions.all().map((action: any) => action.component())}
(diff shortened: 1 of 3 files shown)


components/cms/Toolbar.tsx
```diff
@@ -27,15 +27,25 @@ export const Toolbar = styled(({ ...styleProps }) => {
...
+              const id = `${i}-${git.name}`
+              return git.component(id)
+            })}
         </Github>
         <Status>
           {status.all().length > 0 &&
-            status.all().map((status: any) => status.component())}
+            status.all().map((status: any, i) => {
+              const id = `${i}-${status.name}`
+              return status.component(id)
```
ncphillips: Maybe try this:
```suggestion
              return <status.component key={id} />
```
then the plugin components don't need to set the key themselves

kendallstrautman: nice!

kendallstrautman: Thank you :)

components/cms/Toolbar.tsx
```diff
@@ -27,15 +27,26 @@ export const Toolbar = styled(({ ...styleProps }) => {
           <CreateContentMenu />
         </Create>
         <Github>
-          {git.all().length > 0 && git.all().map((git: any) => git.component())}
+          {git.all().length > 0 &&
+            git
+              .all()
+              .map((git: any, i) => <git.component key={`${i}-${git.name}`} />)}
```
ncphillips: On second look I would also do this:

```suggestion
              .map((git: any, i) => <git.component key={git.name} />)}
```

Using the array index in the `key` is [generally not advised](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318)

Also, `name` is unique across plugins so you'll never have two plugins of the same type with the same.

kendallstrautman: okay, I wasn't sure if there would be multiple instances of a single plugin

PR #202, opened 2020-03-06
Fix typos

DirtyF:
- GitHub
- excited
- Take a peek


DirtyF:
Not what I was expecting, I edited the community page with Tina and hit save, and all I got is a PR with a previous test on the homepage 😄 

/cc @jamespohalloran

PR #205, opened 2020-03-09
handles invalid auth (e.g. revoked app access)

jhuggett:
- handles 401 unauthorized get content error
- when an error is flagged in the modal it clears the preview cookies so that on refresh if the user doesn't follow the flow it kicks em back to a normal viewing mode (also makes it easy to add a cancel auth button in the future (just refresh page)).

### DIFF
--- a/utils/github/getDecodedData.ts
+++ b/utils/github/getDecodedData.ts
@@ -11,11 +11,21 @@ const getDecodedData = async (repoFullName, headBranch, path, accessToken) => {
     path,
     accessToken
   )
-
-  if ((response?.response?.status || 0) == 404) {
-    throw new ContentNotFoundError(
-      'Content not found. Your fork may have been deleted.'
-    )
+  
+  const errorStatus = response?.response?.status || 200
+  if (errorStatus != 200) {
+    switch (errorStatus) {
+      case 404: {
+        throw new ContentNotFoundError(
+          'Content not found. Your fork may have been deleted.'
+        )
+      }
+      case 401: {
+        throw new ContentNotFoundError(
+          'Authentication invalid. You will need to re-authenticate.'
+        )
+      }
+    }
   }
 
   return { ...data, content: b64DecodeUnicode(data.content) }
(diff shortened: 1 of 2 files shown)


utils/github/getDecodedData.ts
```diff
@@ -11,11 +11,21 @@ const getDecodedData = async (repoFullName, headBranch, path, accessToken) => {
...
+  const errorStatus = response?.response?.status || 200
+  if (errorStatus != 200) {
+    switch (errorStatus) {
+      case 404: {
+        throw new ContentNotFoundError(
+          'Content not found. Your fork may have been deleted.'
+        )
+      }
+      case 401: {
+        throw new ContentNotFoundError(
```
jamespohalloran: We should have an UnauthenticatedError here instead of ContentNotFoundError

utils/github/getDecodedData.ts
```diff
@@ -11,11 +11,21 @@ const getDecodedData = async (repoFullName, headBranch, path, accessToken) => {
     path,
     accessToken
   )
-
-  if ((response?.response?.status || 0) == 404) {
-    throw new ContentNotFoundError(
-      'Content not found. Your fork may have been deleted.'
-    )
+  
+  const errorStatus = response?.response?.status || 200
```
jamespohalloran: I ran into an issue last week where we were receiving a successful 201 code, but we were treating it as an error. We may want to abstract this into a helper: https://github.com/tinacms/tinacms.org/blob/master/open-authoring/github/api/index.js#L103

utils/github/getDecodedData.ts
```diff
@@ -11,11 +11,21 @@ const getDecodedData = async (repoFullName, headBranch, path, accessToken) => {
...
+      case 404: {
+        throw new ContentNotFoundError(
+          'Content not found. Your fork may have been deleted.'
+        )
+      }
+      case 401: {
+        throw new ContentNotFoundError(
+          'Authentication invalid. You will need to re-authenticate.'
+        )
+      }
```
jamespohalloran: Maybe we can do a catchall for other types of errors as well with a generic "failed to load content"

PR #206, opened 2020-03-09
Error interpreter

jhuggett:


open-authoring/OpenAuthoringErrorInterpreter.ts
```diff
@@ -0,0 +1,122 @@
...
+            { 
+                message: "Cancel",
+                action: Actions.refresh
+            }], // Action buttons
+            false // clear preview cookies
+        )
+    }
+    
+    switch (parseInt(error.code.toString()[0])) {
+        case 4: {
```
jamespohalloran: I don't quite understand the distinction between 4** (client-side) errors are and 5** (server-side) errors here

jhuggett: It's half because this keeps the functions lean and more readable and the other half because we may do specific checks based on who's having the error (since 4** is supposed to be cs and 5** ss)

components/layout/OpenAuthoringSiteForm.tsx
```diff
@@ -158,22 +162,43 @@ const OpenAuthoringSiteForm = ({
           )}`
         ),
       afterSubmitFailed: failedForm =>
-        setStatefulPreviewError(failedForm.getState().submitError),
+        setStatefulError(failedForm.getState().submitError),
```
jamespohalloran: We're currently keeping 2 errors in state: "statefulError", and "interpretedError". Can we move the logic from useEffect here, and just setState the `await interpretError(statefulError)` result here?

jamespohalloran: Also, we only really need to keep `asModal` in state. Alerts are fire and forget.

open-authoring/OpenAuthoringContextualError.ts
```diff
@@ -0,0 +1,18 @@
+import { Actions } from "./OpenAuthoringModalContainer"
+
+export default class OpenAuthoringContextualError extends Error {
+    title: string
+    actions: [{message: string, action: Actions}]
+    shouldClearPreview: boolean
```
jamespohalloran: Can clearing the previewData just be in the actions?

open-authoring/OpenAuthoringErrorInterpreter.ts
```diff
@@ -0,0 +1,122 @@
...
+import OpenAuthoringContextualError from "./OpenAuthoringContextualError";
+import { Actions } from "./OpenAuthoringModalContainer";
+import { isForkValid } from "./github/api";
+
+export default async function interpretError(error: OpenAuthoringError) : Promise<OpenAuthoringContextualError> {
+    if (!error || !error.code) {
+        return new OpenAuthoringContextualError(
+            false, // should it be presented as a modal? (if not present a toast)
+            "Error", // title
+            "Failed to interpret error. No context provided.", // message (the only thing a toast will present)
```
jamespohalloran: This message feels a bit more like what a developer should see instead of the error message the user will see. Maybe something more generic?

open-authoring/OpenAuthoringErrorInterpreter.ts
```diff
@@ -0,0 +1,122 @@
...
+        }
+        case 5: {
+            return interpretServerError(error)
+        }
+    }
+
+    return new OpenAuthoringContextualError(
+        false,
+        "Error",
+        "Failed to interpret error.",
```
jamespohalloran: Also feels a bit developer'ey for a user error message

open-authoring/OpenAuthoringErrorInterpreter.ts
```diff
@@ -0,0 +1,122 @@
...
+    )
+}
+
+function interpretServerError(error: OpenAuthoringError) {
+    switch (error.code) {
+        case 500: {
+            return new OpenAuthoringContextualError(
+                true,
+                "500 Error",
+                "Server error. Your fork may be out of sync.",
```
jamespohalloran: I almost feel like the message here should be pulled off of the error itself

jamespohalloran: Same with title, and all the non-useraction data

open-authoring/OpenAuthoringErrorInterpreter.ts
jhuggett: those messages are more dev'ey to let us know that we have an error that isn't being interpreted properly, but we can certainly make it more generic / understandable to be user friendly

open-authoring/OpenAuthoringErrorInterpreter.ts
```diff
@@ -0,0 +1,122 @@
+import OpenAuthoringError from "./OpenAuthoringError";
+import OpenAuthoringContextualError from "./OpenAuthoringContextualError";
+import { Actions } from "./OpenAuthoringModalContainer";
+import { isForkValid } from "./github/api";
+
+export default async function interpretError(error: OpenAuthoringError) : Promise<OpenAuthoringContextualError> {
```
jamespohalloran: I think it feels a bit weird to me that this function is designed to return an error instance. Maybe instead of an error object, it's an "OpenAuthoringErrorUI" object that doesn't extend Error? 
If we want to treat OpenAuthoringContextualError as an actual error, I would expect the behaviour to catch a generic "OpenAuthoringError", and throw a more specific OpenAuthoringContextualError

open-authoring/OpenAuthoringErrorInterpreter.ts
jamespohalloran: This is shown to the user in the modal though right? Ideally I'd say we'd want to do a console.warn for the developer message and then something more generic in the modal

PR #210, opened 2020-03-09
Update introducing-visual-open-authoring.md

kendallstrautman:


spbyrne:
I think these changes are covered in #209

PR #271, opened 2020-03-19
fix(errors): fix issue where errors not parsed as JSON

jamespohalloran:
This error surface with the upgrade to Next 9.3.0. Causes the Error modals to not be rendered in preview-mode


pages/blog/[slug].tsx
```diff
@@ -126,7 +126,7 @@ export const getStaticProps: GetStaticProps = async function({
     )
   } catch (e) {
     if (e instanceof OpenAuthoringError) {
-      previewError = e
+      previewError = { ...e } //workaround since we cant return error as JSON
```
Comment (ncphillips) [inline pages/blog/[slug].tsx]: nice fix

PR #279, opened 2020-03-20
GitHub plugin

jamespohalloran:
Pulling out Github specific logic into it's own directory.

This PR accomplishes a few things:
**Client-side:**
- All of the client-side github related functionality uses the Github proxy, so that behavior has been pulled into [GithubAPI](https://github.com/tinacms/tinacms.org/pull/279/files#diff-7f2df370df2837ba346b2e77a16774e3).
- The error-handling was inconsistent with our Github API helpers (sometimes swallowing errors, somethings throwing, sometimes returning), but now it is fairly consistent.

**Server-side**
- The logic behind API functions has been pulled out, so sites can simply create the properly named API files, and export from our package. I.e [create-github-access-token](https://github.com/tinacms/tinacms.org/pull/279/files#diff-147d87cba222243415a9d6c3ef12a63f)
- The data for the Github Proxy no longer needs to be nested in a `proxy_data` property 

Since this change, passing the GitHub provider through the error interpreter's has become a bit smellier. I think this should be resolved with a different PR with a more substantial refactor of the error-handlers.

### DIFF
--- a/open-authoring/error-interpreter/client-side/index.ts
+++ b/open-authoring/error-interpreter/client-side/index.ts
@@ -1,14 +1,19 @@
-import OpenAuthoringError from "../../OpenAuthoringError"
-import interpretUnauthorizedError from "./unauthorized"
-import interpretNotFoundError from "./notFound"
+import OpenAuthoringError from '../../OpenAuthoringError'
+import interpretUnauthorizedError from './unauthorized'
+import interpretNotFoundError from './notFound'
 
-export default async function interpretClientError(error: OpenAuthoringError) {
-    switch (error.code) {
-        case 401: { // Unauthorized
-            return interpretUnauthorizedError(error)
-        }
-        case 404: { // Not Found
-            return await interpretNotFoundError(error)
-        }
+export default async function interpretClientError(
+  error: OpenAuthoringError,
+  github: any
+) {
+  switch (error.code) {
+    case 401: {
+      // Unauthorized
+      return interpretUnauthorizedError(error)
     }
+    case 404: {
+      // Not Found
+      return await interpretNotFoundError(error, github)
+    }
+  }
 }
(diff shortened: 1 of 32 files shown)


open-authoring/error-interpreter/client-side/index.ts
```diff
@@ -1,14 +1,19 @@
...
+  switch (error.code) {
+    case 401: {
+      // Unauthorized
+      return interpretUnauthorizedError(error)
     }
-}
\ No newline at end of file
+    case 404: {
+      // Not Found
+      return await interpretNotFoundError(error, github)
```
ncphillips: this makes me feel like maybe something here should actually be in `cms.api.github`...but I'm not exactly sure how

open-authoring/error-interpreter/client-side/notFound.ts
```diff
@@ -1,12 +1,13 @@
...
+import { enterAuthFlow, refresh } from '../actions'
+import { getForkName, getHeadBranch } from '../../utils/repository'
 
 export default async function interpretNotFoundError(
-  error: OpenAuthoringError
+  error: OpenAuthoringError,
+  github: any
 ) {
-  if (await isForkValid(null)) {
+  if (await github.getBranch(getForkName(), getHeadBranch())) {
```
ncphillips: Hmmm, just spitballing. Maybe we're missing some kind of repository object?

```
if (await repository.getBranch())
```

And the repository knows about the github API 🤔

open-authoring/error-interpreter/client-side/index.ts
jamespohalloran: I haven't really touched any of the error-interpreter stuff in this, as I'm thinking it may be a bit too big of a step.

PR #285, opened 2020-03-24
Error refactor

jamespohalloran:
Cleaning up some aspects of error handling:

- [x] ModalContainer contained a lot of openAuthoring specific logic
- [x] Multiple ModalProviders
- [x] error interpreter actions returning a boolean determining if modal should be closed
- [x] OpenAuthoringErrorProps is the same as ActionableModalOptions
- [x] error interpreter requires "github" object get passed through a few layers

The next steps will be cleaning up a few places that react to "OpenAuthoringError" being thrown. It's mostly in getStaticProps so a bit out of scope for this PR.


components/layout/OpenAuthoringSiteForm.tsx
```diff
@@ -16,9 +18,12 @@ const OpenAuthoringSiteForm = ({ form, editMode, path, children }: Props) => {
   // Persist pending changes to localStorage
   useLocalStorageCache(path, form, editMode)
 
+  useUpdateAuthOnFormFail(form)
+
   return (
     <>
       <FormAlerts form={form} />
+      <AutoAuthModal />
```
ncphillips: What do you think of moving `useUpdateAuthOnFormFail(form)` inside `AutoAuthModal`?

```
<AutoAuthModal form={form} />
```

open-authoring/AutoAuthModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+          {
+            name: 'Cancel',
+            action: cancelAuth,
+          },
+        ],
+      })
+    }
+  }, [])
+
+  return <>{errorModalProps && <ActionableModal {...errorModalProps} />}</>
```
ncphillips: Instead of keeping the props in state, we could just have an `errorModal` boolean.

```
const openAuthoring = useOpenAuthoring()

  const [errorModal, setErrorModal] = useState(false)

  const cancelAuth = () => {
    window.history.replaceState(
      {},
      document.title,
      window.location.href.split('?')[0] //TODO - remove only autoAuth param
    )
    setErrorModal(false)
  }

  const showErrorModal = useEffect(() => {
    if (window.location.href.includes('autoAuth')) setErrorModal(true)
  }, [])

  return (
    <>
    {errorModal && 
       <ActionableModa 
        title='Authentication',
         // etc ...
        />
      }
    </>
  )
```

open-authoring/error-interpreter/github-interpeter.ts
```diff
@@ -0,0 +1,64 @@
...
+    case 500: {
+      return {
+        title: 'Error 500',
+        message: error.message,
+        actions: [reauthenticateAction],
+      }
+    }
+  }
+
+  console.warn('Error Interpreter: Could not interpret error ' + error.code)
```
ncphillips:
```suggestion
```

components/layout/OpenAuthoringSiteForm.tsx
jamespohalloran: IMO they are pretty independent. 
We currently keep a semi-recent auth-state in the root OpenAuthoring component. This state determines if we need to re-authenticate the user / create a new fork when they click edit (which pops the window).

useUpdateAuthOnFormFail triggers the root to re-perform that check, so that when the user clicks "continue", if required, we can open the auth window synchronously (otherwise the window gets popup-blocked).

AutoAuthModal grabs a queryParm and shows a modal kicking off the auth-flow on intiialLoad. If we removed this, we'd still want "useUpdateAuthOnFormFail" to do it's thing when a runtime error shows up.

jamespohalloran: I think I confused things by having `errorModalProps` as a variable name in AutoAuthModal (copy-paste mistake)

PR #288, opened 2020-03-25
Open authoring organization

jamespohalloran:
Reorganizing Open-authoring into potential "packages" for when it's moved into Tina. 
_I think_ this covers the open-authoring specific logic apart from:
- Loading content (from github/file-system)
- Tina forms such as _useGithubJsonForm_

Current areas grouped in "open-authoring" are:
- **alerts**
Display modals with prompts when initialContent fails to load, or actions fail due to the Github-state
- **auth**
Helper for popping up windows for authenticating with Github, and creating a fork
- **error**
Helper for generating the UI that gets shown in a modal based on the Github request error.
- **form** (Not sure what to name this yet.)
Helper for re-checking the "open-authoring" state, when errors occur in the form
- "**open-authoring**" (Not sure what to name this yet.)
Manages the state for if we are connected to Github / if fork exists.
- **toolbar plugins**
Basic plugins used in open-authoring workflow. Once we abstract the toolbar, things like form-state won't have to be apart of this group.

Some of these packages are a bit scoped to Github, and probably should be named appropriately. 

There is also the:
**github-api tina plugin**, &
**browser-storage-api** tina plugin

I branched off of this branch and took things further within the `auth-refactor` branch
https://github.com/tinacms/tinacms.org/compare/open-authoring-organization...auth-refactor


jamespohalloran:
Closing for now, as `auth-refactor` is more up to date: https://github.com/tinacms/tinacms.org/compare/master...auth-refactor

Also see: https://github.com/tinacms/rfcs/pull/6#pullrequestreview-383018810

PR #290, opened 2020-03-26
Refactor: Remove local inline components

kendallstrautman:
- [x] The focus ring is not showing up anymore for any of the inline components, something to look into.
- [x]  Also need to address [this issue](https://github.com/tinacms/tinacms/issues/941)


spbyrne:
It looks like the focus ring issue will be solved by the refactor that makes global styles available without the sidebar.

PR #295, opened 2020-03-30
changed validation to on click

jhuggett:


open-authoring/open-authoring/OpenAuthoringProvider.tsx
```diff
@@ -113,55 +101,83 @@ export const OpenAuthoringProvider = ({
...
-        {
-          name: 'Cancel',
-          action: close,
-        },
-        {
-          name: 'Continue to GitHub',
-          action: async () => {
-            await openAuthoring.authenticate()
-            onUpdateAuthState()
+  openAuthoring.userIsAuthenticated().then( isTrue => {
```
jamespohalloran: "isTrue" var name is pretty ambiguous. Maybe "isAuthenticated"?

jamespohalloran: Also, can we use await throughout this function instead of "then"

open-authoring/open-authoring/OpenAuthoringProvider.tsx
```diff
@@ -113,55 +101,83 @@ export const OpenAuthoringProvider = ({
...
-        {
-          name: 'Cancel',
-          action: close,
-        },
-        {
-          name: 'Continue to GitHub',
-          action: async () => {
-            await openAuthoring.authenticate()
-            onUpdateAuthState()
+  openAuthoring.userIsAuthenticated().then( isAuthenticated => {
```
jamespohalloran: I believe this will make a request every time this modal rerenders

open-authoring/open-authoring/OpenAuthoringProvider.tsx
```diff
@@ -113,55 +101,83 @@ export const OpenAuthoringProvider = ({
   )
 }
 
-const OpenAuthoringAuthModal = ({ onUpdateAuthState, close }) => {
-  let modalProps
+const OpenAuthoringAuthModal = ({ close }) => {
+  const [modalProps, setModalProps] = useState({
```
jamespohalloran: It feels a bit weird to me to have the loading state be using these modal props.

open-authoring/open-authoring/OpenAuthoringProvider.tsx
jamespohalloran: I think this could maybe be reorganized a bit. We already made a request to see if we are authorized and if the fork exists, to get to the point where we're showing the modal. Can we pass those values through as props? Also, then I don't think we would even need a loading state

PR #296, opened 2020-03-30
Toolbar Refactor

ncphillips:
This PR implements the changes described in [RFC 0005](https://github.com/tinacms/rfcs/blob/master/0005-toolbar-widget.md)

* The `toolbar:form-action` and `toolbar:status` plugins have been removed. Those components are always rendered now. This is the same way it works in the Sidebar.
* The `toolbar:git` plugin has been replaced with a more generic `toolbar:widget`. 
  * These widgets are ordered by `weight`.
  * Any `props` defined on the widget are passed to the component.

### DIFF
--- a/components/cms/Toolbar.tsx
+++ b/components/cms/Toolbar.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useEffect } from 'react'
-import { useCMS, useSubscribable, Form, FieldMeta } from 'tinacms'
+import { useCMS, useSubscribable, Form, FieldMeta, Plugin } from 'tinacms'
 import { Button, color } from '@tinacms/styles'
 import { CreateContentMenu } from './CreateContent'
 import styled, { css } from 'styled-components'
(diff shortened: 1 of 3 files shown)


components/cms/Toolbar.tsx
```diff
@@ -27,23 +53,39 @@ export const Toolbar = styled(({ ...styleProps }) => {
           <CreateContentMenu />
         </Create>
         <Github>
```
jamespohalloran: Maybe out of scope for this PR, but this doesn't seem like it needs to be scoped to Github

components/cms/Toolbar.tsx
```diff
@@ -27,23 +53,39 @@ export const Toolbar = styled(({ ...styleProps }) => {
...
-              .all()
-              .map((action: any) => <action.component key={action.name} />)}
-        </Actions>
+        {formState && (
+          <>
+            <Status>
+              <FormStatus dirty={!formState.pristine} />
+            </Status>
+            <Actions>
+              <ToolbarButton disabled={formState.pristine} onClick={form.reset}>
```
jamespohalloran: This breaks on the the blog list page with:
```
TypeError: Cannot read property 'reset' of null
```

PR #298, opened 2020-03-30
added validation on click and fixed checks

jhuggett:


open-authoring/open-authoring/OpenAuthoringProvider.tsx
```diff
@@ -57,11 +57,26 @@ export const OpenAuthoringProvider = ({
   const [authorizing, setAuthorizing] = useState(false)
 
   const updateAuthChecks = async () => {
-    setAuthenticated(!!(await cms.api.github.getUser()))
-    setForkValid(await cms.api.github.getBranch(getForkName(), getHeadBranch()))
+    const user = await cms.api.github.getUser()
+    if (user) {
+      setAuthenticated(true)
+      const forkName = getForkName() || (user.login || "") + "/" + process.env.REPO_FULL_NAME.split("/")[1]
```
jamespohalloran: I'm a bit torn on this optimization. It increases the complexity of this function a bit, and I'm not sure that it will scale once we add more use-cases (forking to an organization, accessing an other user's fork, etc)

open-authoring/open-authoring/OpenAuthoringProvider.tsx
```diff
@@ -80,13 +95,6 @@ export const OpenAuthoringProvider = ({
     }
   }, [authorizing, forkValid, authenticated])
 
-  // Hook to update root openAuthoring state when form fails.
-  // We need to perform to check before an action is clicked (e.g start auth flow)
-  // Because if it is perform on-the-fly, the window may be blocked.
-  useEffect(() => {
```
jamespohalloran: I think we'll still want this, it's for when we show an error-modal we will want to get the latest authentication state

open-authoring/github-auth/validation.ts
```diff
@@ -0,0 +1,27 @@
+import { getForkName, getHeadBranch, setForkName } from "../open-authoring/repository"
+import { useCMS } from "tinacms"
+
+export const validate = async () => {
+    var validations = {
+        userIsAuthenticated: false,
+        forkIsValid: false
+    }
+
+    const cms = useCMS()
```
jamespohalloran: hooks typically shouldn't be called in functions. Depending on how the function is called it could violate the rules of hooks and blow up: https://reactjs.org/docs/hooks-rules.html

PR #329, opened 2020-04-07
Docs: Update Inline Editing

kendallstrautman:
## What this does:

- Updates the general guidelines for adding Inline Editing
- Adds docs for Inline Fields
- Removes old gatsby and next.js-specific inline editing docs (adds redirects)

These docs feel super similar to the blocks field docs...but not quite sure the best way to remedy this. Open to feedback!

**Still Left:** 
- [x] Need to make sure redirects for old inline docs are working
- [x] Proofread & Edit


kendallstrautman:
@DirtyF - redirects should work now 👌🏼

PR #340, opened 2020-04-15
[DRAFT] Prototype: Branch Switcher UI

spbyrne:
Creating this PR so we can have a live preview of what I'm working on.


DirtyF:
@spbyrne can't see it, I got stuck on the GitHub Authorisation part.

spbyrne:
Unfortunately it seems like the auth only works on the live domain or in testing, so to demo this we'd need to merge.

PR #361, opened 2020-04-21
Upgrade tinacms GitHub

### LINKED ISSUE #356: Update tinacms.org to use getGithubPreviewProps
Our layouts on tinacms.org should switch to use [getGithubPreviewProps](https://github.com/tinacms/tinacms/blob/master/packages/next-tinacms-github/src/get-github-preview-props.ts) helper for grabbing / formatting its Github content

jamespohalloran:
Upgrade "tinacms", "react-tinacms-github", and "next-tinacms-github". Switch up the layouts to match the data loaded from getGithubPreviewProps

tinacms upgrade fixes toolbar dirty-state issues

fixes #356


components/layout/Layout.tsx
```diff
@@ -12,15 +12,15 @@ interface LayoutProps {
...
   sourceProviderConnection: any
-  editMode: boolean
+  preview: boolean
 }
 
 export const Layout = styled(
   ({
     children,
     color,
     sourceProviderConnection,
```
ncphillips, line 22: I bet this can be removed

pages/blog/[slug].tsx
```diff
@@ -48,12 +36,12 @@ function BlogTemplate({
   return (
     <OpenAuthoringSiteForm
       form={form}
-      path={markdownFile.fileRelativePath}
-      editMode={editMode}
+      path={file.fileRelativePath}
```
Comment (ncphillips) [inline pages/blog/[slug].tsx:39]: This makes me want to rename that property to `file.relativePath`
```
      path={file.fileRelativePath}
```

components/layout/Layout.tsx
ncphillips: Just checked. `sourceProviderConnection` is only being used to passed to the CreateBlogPostPlugin. That plugin no longer needs `sourceProviderConnection` so we can remove all these references 🙌

PR #377, opened 2020-04-24
Adds Toolbar Docs

kendallstrautman:
- removed outdated `theme` property on sidebar config
- added the toolbar state options to the examples and interface

I think there's room to expand on Toolbar usage in an Inline Editing guides, but for the base documentation I think this works. Perhaps I need to add a sentence as to what the toolbar is?


dwalkr:
What you have here is good for the sake of accuracy/completeness in the CMS config. I think we could use some documentation elsewhere that discusses the different main UI elements and explain the toolbar there. I'd hesitate to include Toolbar explanation within the Inline Editing docs because its scope is a bit broader than that.

PR #345, opened 2020-04-16
Remove useLocalForms references from docs

ncphillips:

PR #385, opened 2020-04-27
Adds Next-Git Guide & General Next.js package docs

kendallstrautman:
Converts (almost verbatim) the old next.js docs to a 'Git-based' guide. Pulls over initial changes from #345 to deprecate local/global helpers. I took out all references to global helpers. I think this should just be on the main _Forms_ docs since its pretty agnostic to the data source.


kendallstrautman:
This 'guide' feels like a hybrid between a guide and documentation. For example with all the form helpers. I'm wondering if I should delete the robust form helper info from the guide and put that in a Next+Git doc where it can be more bare bones...

kendallstrautman:
I reorganized things a bit so that there are core next.js docs. I trimmed out the superfluous form helper info from the next-git guide. 

I think the doc that needs the most work is the `next-tinacms-github` one. It should do a better job of listing the helpers, interfaces and such.

dangayle:
I tried this with a private Github repo, it failed and asked me if it were public. Could that be documented and/or make it so that my repo can be private?

kendallstrautman:
Hey @dangayle, appreciate the feedback! The guide on this PR is actually the _Git_ guide, and not the _GitHub Open Authoring_ guide. However, I do believe it is mentioned in the very first steps of that guide that repositories need to be public with Open Authoring. I could add some more tips around that if it seems to be a common stumbling block. 

Regarding your fork failing, I would suggest double checking the _env_ values match the OAuth app, and sometime you need to hit the 'create fork' button twice I have noticed. For follow up questions or comments, the forum would be a great place to get help on this! If you find a bug or _issue_ with the setup, feel free to create an issue in the `tinacms/tinacms` repo.

ncphillips:
What's the state of this?

kendallstrautman:
It just needs to be reviewed I think.

PR #417, opened 2020-05-12
Update plugin docs to reflect module syntax and async plugin registration

chrisdmacrae:
Please pull these awesome changes in!


content/docs/cms/plugins.md
```diff
@@ -30,38 +29,62 @@ Call `cms.plugins.add` and pass in the plugin.
...
+  
+  return <div>{children}</div>
 }
 ```
 
 When adding plugins from inside a React component, the plugin is added when the component mounts, and removed when the component unmounts. This is both expected and encouraged, as Tina has a [Dynamic Plugin System](/blog/dynamic-plugin-system).
 
+### Adding Plugins Dynamically
+
+In some cases, you may not want plugins to be included in the initial javascript file/bundle of your website, as some plugins can be quite large. In these scenarios, you can import the plugins dynamically, and the plugins will only be downloaded when they are needed.
```
DirtyF, line 58:
```suggestion
In some cases, you may not want plugins to be included in the initial JavaScript file/bundle of your website, as some plugins can be quite large. In these scenarios, you can import the plugins dynamically, and the plugins will only be downloaded when they are needed.
```

ncphillips:
Must wait for tinacms/tinacms#1134 to be merged and deployed.

PR #432, opened 2020-05-26
"inline editing" docs updates

ncphillips:
Remove references to `useLocalForm`

### DIFF
--- a/content/docs/inline-editing/inline-wysiwyg.md
+++ b/content/docs/inline-editing/inline-wysiwyg.md
@@ -27,7 +27,7 @@ import { InlineForm, InlineWysiwyg } from 'react-tinacms-inline'
 
 // Example 'Page' Component
 export function Page(props) {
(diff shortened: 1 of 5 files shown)


content/docs/inline-editing/inline-wysiwyg.md
```diff
@@ -23,12 +22,13 @@ Below is an example of how an `InlineWysiwyg` field could be defined in an [Inli
 
 ```jsx
 import ReactMarkdown from 'react-markdown'
-import { useLocalForm } from 'tinacms'
+import { useForm, usePlugin } from 'tinacms'
 import { InlineForm, InlineWysiwyg } from 'react-tinacms-inline'
 
 // Example 'Page' Component
 export function Page(props) {
   const [data, form] = useLocalForm(props.data)
```
ncphillips, line 30:
```suggestion
  const [data, form] = useForm(props.data)
```

PR #433, opened 2020-05-27
Blocks & Inline docs updates

kendallstrautman:
Now that individual[ block fields are deprecated](https://github.com/tinacms/tinacms/pull/1160), this PR condenses the blocks docs into a single file and references the other inline fields to be used within blocks. It also [updates the names](https://github.com/tinacms/tinacms/pull/1176) of the inline fields and adds the `focusRing` option to the inline field interfaces. Also this [fixes the import info](https://github.com/tinacms/tinacms/pull/1134) for `InlineWysiwyg`. Adds `InlineGroup` docs!


ncphillips:
This looks great. Just gonna wait to approve until Monday

PR #436, opened 2020-05-28
fix: updated inline-wysiwyg docs

logan-anderson:
got an error when I did
```
yarn add react-tinacms-{editor,github,inline}@next
```
I fixed it by switching the import to this. Should we update docs?


kendallstrautman:
Yep! Just updated it in #433 but we can merge this in now since we'll wait for the next release with that PR👌🏼

PR #442, opened 2020-06-03
docs: updates inline group docs with latest changes

kendallstrautman:


kendallstrautman:
I'd like to remove the focus ring info from this PR before it gets merged. Changes have already been made to that API

PR #445, opened 2020-06-08
fix: typos on release notes

DirtyF:
Awesome release, congrats to the team.

_Edited with Tina 🦙_

### DIFF
--- a/content/blog/2020-06-08-release-notes.md
+++ b/content/blog/2020-06-08-release-notes.md
@@ -5,13 +5,13 @@ author: Nolan Phillips
 ---
 We missed release notes for the June 1st release so this week's notes are packed!
 
-There's been a lot of work on the Inline Editing experience in the past couple weeks. The most notable changes are
+There's been a lot of work on the Inline Editing experience in the past couple of weeks. The most notable changes are:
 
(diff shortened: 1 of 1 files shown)


content/blog/2020-06-08-release-notes.md
```diff
@@ -86,9 +86,9 @@ Thanks to everyone for contributing!
 | 12 | Brent Mitton |
 | 3 | Logan Anderson |
 | 3 | Sean MacKay |
-| 2 | jpuri |
-| 1 | DJ |
+| 2 | Jyoti Puri |
+| 1 | DJ Walker |
```
ncphillips, line 91: There's no name convention being used here. The names listed are the names that the developer signed the commit with.

DirtyF: Right, I just figured, I might as well name our peers properly while at it.

ncphillips: Yeah I'm okay with the change. I'm just noting that this is the name _they_ chose to set in their git config so I don't modify it in release notes. It's not my responsibility to name people. Sometimes people intentionally don't add their full/real name for privacy reasons.

PR #458, opened 2020-06-10
Update inline docs & cms.enabled

kendallstrautman:
- Adds `focusRing` docs 
- Adds `direction` docs for `InlineBlocks`
- Adds `insetControls` prop to `InlineBlocks` and `InlineGroup`
- Other interface corrections for inline fields
- Adds `enabled` to cms docs and info on how to use methods
- Removes references to setting inline edit mode with methods from `useInlineForm`, adds note on suggestion to use Toolbar for these controls.

Could use more examples on how all these things can play together but I'm thinking that will be better in the guide.


content/docs/cms.md
```diff
@@ -92,12 +92,33 @@ export default function SomeComponent() {
...
+```js
+import * as React from 'react'
+import { useCMS } from 'tinacms'
+
+// <ExitButton /> is assumed to be nested inside of a <Tina> component
+export default function ExitButton() {
+  const cms = useCMS()
+
+  return (
+    <button onClick={cms.enabled ? cms.disable() : cms.enable()}>
```
ncphillips, line 108: Every time I look at this I think that maybe there should just be a `cms.toggle()` function as well

content/docs/inline-editing/inline-blocks.md
```diff
@@ -58,6 +58,32 @@ interface BlockComponentProps {
...
 
+#### Blocks Controls Options
+
+| Key             | Description                                                                                                                                                                                                                                                                                                                        |
+| --------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `name`          | The path to some value in the data being edited. If no value is provided, the child fields will reference the root of the source file.                                                                                                                                                                                             |
```
ncphillips, line 65: `BlocksControls` doesn't have a `name` prop, it accesses it via the `InlineFieldContext`
```suggestion
```

kendallstrautman: good catch, thanks

content/docs/cms.md
kendallstrautman: yessss

ncphillips: https://github.com/tinacms/tinacms/pull/1265

PR #462, opened 2020-06-12
Release notes

ncphillips:


ncphillips:
@dwalkr or @kendallstrautman 

The `src` of the `<iframe>` in the Markdown body gets mangled. I can switch to an image/link if necessary

image

kendallstrautman:
Yikes, Is that Tina's doing? Can we just manually write the src?

ncphillips:
It's not Tina doing it it's `react-markdown`

ncphillips:
Wait no you were right, it _was_ the wysiwyg

PR #466, opened 2020-06-15
updated date field docs

ncphillips:
- remove gatsby from example
- switch from options list ot table
- add dateFormat and timeFormat to options table

### DIFF
--- a/content/docs/fields/date.md
+++ b/content/docs/fields/date.md
@@ -13,16 +13,50 @@ The `date` field represents a date and time picker. It can be used for data valu
 
 ![tinacms-date-field](/img/fields/date.jpg)
 
-## Definition
+## Options
 
-Below is an example of how a `date` field could be defined in a Gatsby remark form. Read more on passing in form field options [here](/docs/gatsby/markdown#customizing-remark-forms).
+This field plugin uses [`react-datettime`](https://www.npmjs.com/package/react-datetime) under the hood.
+
+```typescript
+interface DateConfig extends FieldConfig, DatetimepickerProps {
+  component: 'date'
+  name: string
+  label?: string
+  description?: string
+  dateFormat?: boolean | string // Extra properties from react-datetime
+  timeFormat?: boolean | string // Moment date format
+}
+```
+
+| Option        | Description                                                                                                                                                                                                                                                                                                                                                                                                               |
+| ------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


content/docs/fields/date.md
```diff
@@ -13,47 +13,55 @@ The `date` field represents a date and time picker. It can be used for data valu
...
+| `timeFormat`  | Defines the format for the time. It accepts any [Moment.js time format](https://momentjs.com/docs/#/displaying/format/) (not in localized format). If true the time will be displayed using the defaults for the current locale. If false the timepicker is disabled and the component can be used as datepicker. See [react-datetime docs](https://github.com/YouCanBookMe/react-datetime) for more details._(Optional)_ |
 
-- `name`: The path to some value in the data being edited.
-- `component`: The name of the React component that should be used to edit this field. Available field component types are [defined here](/docs/fields)
-- `label`: A human readable label for the field. This label displays in the sidebar and is optional. If no label is provided, the sidebar will default to the name.
-- `description`: An optional description that expands on the purpose of the field or prompts a specific action.
+> ### FieldConfig
```
logan-anderson, line 40: Thanks for adding this! This confused me at first not knowing that FieldConfig was.

content/docs/fields/date.md
```diff
@@ -13,47 +13,55 @@ The `date` field represents a date and time picker. It can be used for data valu
...
+      label: 'Date',
+      name: 'created_at',
+      component: 'date',
+      dateFormat: 'MMMM DD YYYY',
+      timeFormat: false,
+    },
+    // ...
+  ],
+}
+```
```
logan-anderson, line 67: Would it be worth it also linking to a real-world example like [BlogPostCreator]( https://github.com/tinacms/tinacms.org/blob/e217a45a2408ff6fd713939885f886a342953a40/tinacms/BlogPostCreator.ts#L19)  or [blog slug page](https://github.com/tinacms/tinacms.org/blob/e217a45a2408ff6fd713939885f886a342953a40/pages/blog/%5Bslug%5D.tsx#L165) (tina cms site).  That way folks could see a working example. or would this just cause problems for maintenance?

ncphillips: Hmm 🤔 These docs are meant to be purely Tina (i.e. no reference to frameworks or backends) but I definitely see the value in that. 

What do the other@tinacms/maintainers think?

ncphillips: Shit my mention didn't work.

@tinacms/maintainers

kendallstrautman: Yeah I'd rather just have the examples in this doc. If there are other use cases that should be noted, let's add them!

PR #471, opened 2020-06-16
Adds inline editing guide

kendallstrautman:
## Update 06.18.20 

This is now in a good place for anyone to review.

[This is the demo](https://github.com/tinacms/inline-blocks-demo).

The [`master`](https://github.com/tinacms/inline-blocks-demo) branch can now serve as our 'step 1' to test off of. Make new branches off that to run through the flow. Use the [`final-state`](https://github.com/tinacms/inline-blocks-demo/tree/final-state) branch as a reference. 

- [x] Add more assets
- [x] Set up final repo state and link the 'step 1' commit
- [x] Check for other `TODO` comments
- [x] Update README in demo
- [x] Delete any extra branches in the demo
- [x] Apply edits from all feedback


content/guides/general/inline-blocks/inline-form-fields.md
```diff
@@ -0,0 +1,131 @@
+---
+title: Add an Inline Form & Fields
+---
+
+## Step 2 - Register a generic form
+
+<!-- Add tagged commit -->
+
+The first thing we are going to do is create an _Inline Form_ with [`useForm`](<[https://tinacms.org/docs/forms#creating-forms-in-react](https://tinacms.org/docs/forms#creating-forms-in-react)>). Note that since we don't have a backend set up, the data changes won't persist and the toolbar won't save / track form status as expected. Please refer to [other guides](<[https://tinacms.org/guides/](https://tinacms.org/guides/)>) to learn more on setting up a backend.
```
ncphillips, line 9: Is it just me or are all these links messed up?

kendallstrautman: probably formatting issues from the export

content/guides/general/inline-blocks/hero-to-block.md
```diff
@@ -0,0 +1,126 @@
...
+  InlineTextarea,
+  InlineText,
++  BlocksControls,
+} from 'react-tinacms-inline';
+import '../styles/Hero.css';
+
+-export function Hero() {
++ export function Hero({index}) {
+  return (
++   <BlocksControls index={index}>
```
ncphillips, line 25: What if we didn't make the `BlocksControls` part of the `Hero` component. 

The first `Hero` could be wrapped in an `<InlineGroup name="hero">` instead, and then we could define the `HeroBlock` in the same file:

**components/hero.js**
```tsx
export function Hero() {
  // ...
}

const template = { ... }

export HeroBlock = {
  template, 
  Component(props) {
    return <BlocksControles index={index.index}><Hero {...props} /></BlocksControls>
  }
}
```

The idea being that you can define components with inline fields and use them in blocks, groups, or just straight in the form.

kendallstrautman: I like your thinking, the issue though is that at first there are no fields for the settings modal and the `InlineGroup` is useless without those extra fields. But it could go, make it a group when we add the inline fields, and then add settings modal fields, and _then_ convert to a block...

While I like the diversity in showing how you can define the `Component`, I think the first example should be the most simple and probably the most common way people will define blocks. But perhaps we could touch on this alternative way of defining the component in the later 'Add More Blocks' section?

logan-anderson:
@kendallstrautman  Ran through the first 13 steps of the guide here are my notes

Fist of all awesome work! this is shaping up to be a really good guide. I enjoyed reading it.

- we could link to inline form docs in step 2?
- maybe a quick explanation about how the "data" syncs with the JSON through the name
	-  in step 4 the last paragraph maybe add something like ```in this case it would be data/data.json```
- uppercase H on hero.js in step 6 
- the diff for importing Hero template import needs a ```+``` step 6-7
- Step 10 the brackets on the style tag to not match }} needs to be moved
	-  and on step 10 the H needs to be uppercase on hero
-  assets folder needs to be added (I got this error on step 11)

```
./src/components/Images.js 
Module not found: Can't resolve '../assets/ivan-bandura-unsplash.jpg' in '/home/logan/Documents/inline-blocks-demo/src/components'
```

spbyrne:
These are mainly nits / minor issues

* Inconsistent use of semi colons in example code (fixed by prettier, but only if they have it set up)
* The removed lines in the code diff examples are hard to see (red is too subtle in our code color scheme).
* Turning the hero into a block before we have InlineBlocks set up felt weird, I would expect to do it the other way around. I felt the same when using the inline blocks docs.
* Definitely feeling some info overload with the # of steps & content for each step, but there wasn't a lot of unnecessary information. I may do another read through and see what could potentially get cut down.
* The "Creating Nested Blocks” step tripped me up. Based on previous components steps you’d be tempted to copy/paste the first two code examples unless you read them closely. This may also be the first step with a partial data.json file change, so you have to be careful in what you copy/paste and where.
* No real conclusion / tie up.

content/guides/general/inline-blocks/hero-to-block.md
ncphillips: I just saw this.

What if the first data wasn't grouped under 'block' and you didn't need to use `InlineGroup` at all?

content/guides/general/inline-blocks/inline-form-fields.md
```diff
@@ -0,0 +1,137 @@
...
+    id: './data/data.json',
+    initialValues: {
+      hero: data.hero,
+    },
+    onSubmit() {},
+  }
+
+  // 3. Call `useForm` with the config object.
+  const [pageData, form] = useForm(formConfig)
+
```
ncphillips, line 32:
```suggestion
import { useForm, usePlugin } from 'tinacms'
import { Hero } from './components/Hero'
import data from './data/data.json'

export default function Home() {
  // 2. Create a config object
  const formConfig = {
    id: './data/data.json',
    initialValues: {
      hero: data.hero,
    },
    onSubmit() {
      alert("Saved!")
    },
  }

  // 3. Call `useForm` with the config object.
  const [pageData, form] = useForm(formConfig)
  usePlugin(form)
```

content/guides/general/inline-blocks/overview.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Set-up the demo repo
+
+To get started, you'll need to clone [the repo](https://github.com/tinacms/inline-blocks-demo), install the dependencies. The `master` branch is set-up to be your _initial starting point_.
+
+At this point, it is set up with a TinaCMS instance and renders a simple `Hero` component with hard-coded data. All of the styles and images are configured as well, so when you add blocks they'll look nice 💅.
+
+```bash
+git clone git@github.com:tinacms/inline-blocks-demo.git
+cd inline-editing-demo
```
kendallstrautman, line 21:
```suggestion
cd inline-blocks-demo
```

content/guides/general/inline-blocks/inline-form-fields.md
```diff
@@ -0,0 +1,150 @@
...
+    <div className="hero">
+      <div className="wrapper wrapper--narrow">
+-       <h1>{data.headline}</h1>
++       <h1>
++        <InlineTextarea name="headline" />
++       </h1>
+-       <p>{data.subtext}</p>
++       <p>
++         <InlineTextarea name="subtext" />
++       </p>
```
mittonface, line 135:
```suggestion
+        <InlineTextarea name="data.headline" />
+       </h1>
-       <p>{data.subtext}</p>
+       <p>
+         <InlineTextarea name="data.subtext" />
+       </p>
```

This is the only way this works for me. I _don't think_ I've missed a step or anything.

kendallstrautman: yeah the example is messed up, we're fixing it 👌

PR #482, opened 2020-06-19
add webpack plugins for bundle analyzing and moment locale stripping

ncphillips:
* reduces bundle size by 428kb
* gives us a visual into the websites bundle

[See this comment for more info on bundle size improvements](https://github.com/tinacms/tinacms/issues/771#issuecomment-646668855)

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -12,43 +12,45 @@ const withBundleAnalyzer = require('@next/bundle-analyzer')({
 const dummyMailchimpEndpoint =
   'https://theDomainHere.us18.list-manage.com/subscribe/post?u=1512315231251&amp;id=0asd21t12e1'
 
-module.exports = withBundleAnalyzer(
-  withSvgr({
-    env: {
-      MAILCHIMP_ENDPOINT:
-        process.env.NODE_ENV === 'production'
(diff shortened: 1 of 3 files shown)


next.config.js
```diff
@@ -1,44 +1,54 @@
...
-  exportTrailingSlash: true,
-  exportPathMap: async function() {
-    return {}
-  },
-  webpack(config) {
-    if (process.env.TINA) {
-      let watch
-      if (process.env.TINA_WATCH) {
-        watch = process.env.TINA_WATCH.split(',')
+module.exports = withBundleAnalyzer(
```
logan-anderson, line 15: Looks good! only thing is we might consider doing something like 
```js
const nextconfig = {
   env{
      //...
   }
//...
}
```
and then
```js
export withBundleAnalyzer(withSvgr(nextconfig));
```
might be cleaner if you add more of these nextjs plugins.

you can look at with[ nextjs.org does here](https://github.com/vercel/next-site/blob/3b270c0644c346af832dcbd5a973a1276f0d4335/next.config.js#L198)

We don't need to make that change now just might make things easier later.

ncphillips: Agreed. Might as well do that now.

PR #505, opened 2020-06-25
Events Docs

ncphillips:
[Check them out here](https://tinacms-site-next-git-events-docs.tinacms.vercel.app/docs/cms/events)

### DIFF
--- a/content/docs/cms/events.md
+++ b/content/docs/cms/events.md
@@ -6,4 +6,85 @@ next: /docs/cms/alerts
 
 The **Events** feature allows decouple parts of the CMS to notify the rest of the system about certain events which have happened.
 
-## Events
+## Interface
+
+```ts
+interface Events {
+  dispatch(event: Event): void
+  subscribe(event: string, listener: Listener): Unsubscribe
+}
+
+interface Event {
+  type: string
+  // Any other data may be added to the event.
+  [key: string]: any
+}
+
+type Listener = (event: Event) => void
+
+type Unsubscribe = () => void
+```
+
+## Usage
+
+### Dispatching Events
+
+```ts
+cms.events.dispatch({
+  type: 'something-happened',
+})
+```
+
+### Subscribing to Events
+
+#### Log All Events
+
+```ts
+cms.events.subscribe('*', event => {
+  console.log(event)
+})
+```
+
+#### Log all Plugin Events
+
+```ts
+cms.events.subscribe('plugins', ({ plugin }) => {
+  console.log(`Something happened to the plugins`)
+})
+```
+
+#### Log when a Form Plugin is added
+
+```ts
+cms.events.subscribe("plugins:add:form", ({ plugin }) => {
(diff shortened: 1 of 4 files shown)


content/docs/cms/events.md
```diff
@@ -0,0 +1,90 @@
+---
+title: Events
+prev: /docs/cms/apis
+next: /docs/cms/alerts
+---
+
+The **Events** feature allows decouple parts of the CMS to notify the rest of the system about certain events which have happened.
```
jamespohalloran, line 7:
```suggestion
The **Events** feature allows decoupled parts of the CMS to notify the rest of the system about certain events which have happened.
```

IronSean:
If you subscribe to "pugins:add" will you get all plugin add events, or do you have to do "plugins:add:*"?

"Log all Form Events
cms.events.subscribe("plugins:*:form", ({ plugin }) => {
  console.log(`Added form "${plugin.__type}" called "${plugin.name}"`
})"

Could update the text to not just say added there, since it could be removed as well.

Event can take any shape beyond having a type property, but is there any value in explaining what some of the common TinaCMS shape are, since the types will be generic when developing with them?

IronSean:
`type` on `Event` and `__type` on `Plugin` might be visually similar enough to require some extra thought to parse that they are on different layers of the object and are completely unrelated.

ncphillips:
Thanks for the input @IronSean. I've updated the examples

content/docs/cms/events.md
```diff
@@ -0,0 +1,90 @@
...
+---
+
+The **Events** feature allows decoupled parts of the CMS to notify the rest of the system about certain events which have happened.
+
+## Interface
+
+```ts
+interface Events {
+  dispatch(event: Event): void
+  subscribe(event: string, listener: Listener): Unsubscribe
```
dwalkr, line 14:
```suggestion
  subscribe(eventType: string, listener: Listener): Unsubscribe
```

seeing the two `event` params next to each other with different types makes me think it would be helpful to call this `eventType`

content/docs/cms/events.md
```diff
@@ -0,0 +1,90 @@
+---
+title: Events
+prev: /docs/cms/apis
+next: /docs/cms/alerts
+---
+
+The **Events** feature allows decoupled parts of the CMS to notify the rest of the system about certain events which have happened.
```
dwalkr, line 7:
```suggestion
The **Events** feature allows decoupled parts of the CMS to notify the rest of the system that something has happened.
```

dwalkr: This sentence is kinda heavy to me. I'm not sure if my suggestion makes it better :man_shrugging:

dwalkr:
```suggestion
The **Events** feature allows decoupled parts of the CMS to:

  1. Notify the rest of the system that something has happened, and
  2. React to things that happen elsewhere in the system.
```

PR #506, opened 2020-06-25
feat: added sortable guides

logan-anderson:


logan-anderson:
@ncphillips Made that change

PR #507, opened 2020-06-25
blog: inline editing project recap

kendallstrautman:
[Check it out](https://tinacms-site-next-nqws3qgbi.vercel.app/blog/inline-editing-project)


spbyrne:
This is great! Exactly what we needed, IMO.

DirtyF:
Very good summary of the improvements. 
As I didn't get a chance to play with those yet, I'm curious how nested blocks and inline field settings look, but I guess that will be added in the docs?

kendallstrautman:
You can see a simple example of how nested blocks look and behave on the [hosted demo](https://logan-anderson.github.io/cra-hosted-demo/). Scroll to the bottom 'Features List' type block.

ncphillips:
Once this is published let's add a link to the Release notes

PR #525, opened 2020-07-07
Docs: Remove duplicated words from docs, and remove hard-coded localhost urls from GitClient config docs.

IronSean:
The hard coded localhost url is not necessary as the currently used base url will be used instead (ie, on localhost:3000 it will use localhost:3000, and on 192.168.2.23:3000 it'll use 192.168.2.23:3000), but by hard coding it TinaCMS will not function if the running instance is accessed from a different url. TinaCMS will then become unresponsive and actually lock up the chrome tab if you attempt to access it that way.

Also removed two instances if `are are` in the documentation text.


kendallstrautman:
Thank you! @IronSean

PR #533, opened 2020-07-14
Docs index

ncphillips:
We rewrote the Docs index to serve as a jumping-off point for people visiting the website. This was inspired by the phenomenal [Django Docs](https://docs.djangoproject.com/en/3.0/) as well as the [Rust-lang learn page](https://www.rust-lang.org/learn)

[Checkout the preview here](https://tinacms-site-next-git-docs-index.tinacms.vercel.app/docs)

### DIFF
--- a/content/docs/index.md
+++ b/content/docs/index.md
@@ -33,9 +33,9 @@ Learning how to build a CMS is no small feat! We've compiled a variety of resour
 
 ## Constructing The User Interface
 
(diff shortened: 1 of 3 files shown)


logan-anderson:
Looking really good!!! Bottom nav seems to be off. (Linking to the wrong page and the flow from page to page is off but maybe that's something we can revisit)

PR #536, opened 2020-07-15
Docs & Guides TOC

logan-anderson:
* Added the TOC from the doc starter
* Styled to match new nav
* Highlights current ToC element
* Hidden by default on mobile, click "Show Table of Contents" to expand

Screen Shot 2020-07-22 at 4 23 55 PM


logan-anderson:
@spbyrne I just changed the position. not sure if this was a good change or a bad one.

toc

spbyrne:
Still working on the styles but here's a work-in-progress shot of the desktop and mobile styles:

Screen Shot 2020-07-21 at 4 40 02 PM

Screen Shot 2020-07-21 at 4 41 18 PM

kendallstrautman:
This is awesome! My only feedback is maybe we should only use 'Inter' in the TOC. I find Tuner very hard to read with the longer headings at this small type size.

spbyrne:
> This is awesome! My only feedback is maybe we should only use 'Inter' in the TOC. I find Tuner very hard to read with the at this smaller type size with the long character counts that many docs & guide headings have.

Yeah I think that's a good call, I pulled tuner and it looks better. I adjusted some spacing as well. Some of the text highlighting doesn't work when there's something like italics in the ToC header, but we can probably fix that.

kendallstrautman:
Hmm....I'm seeing a little flickering / re-rendering now. Didn't see that yesterday...Do you get that @spbyrne?

ncphillips:
Confirmed on the flickering. Fixing it could work but also just not having any underlines would be good enough IMO

spbyrne:
Yeah the underline changes are 90% of what bothers me so I'll just remove that

spbyrne:
👍  Thanks all

kendallstrautman:
We need to somehow strip the italics from the anchor tags..... 

https://tinacms.org/guides/general/inline-blocks/inline-form-fields#create-an-_inline-form_ 

doesn't work but this does 

https://tinacms.org/guides/general/inline-blocks/inline-form-fields#create-an-inline-form

I'll make a todo

PR #538, opened 2020-07-15
Doc loading refactor

ncphillips:


logan-anderson:
Looks good! it is working for me and this is a step in the right direction, but why do we need to grab `next` and `prev`? Alternatively, we could generate that by looking at `toc-doc.json`

ncphillips:
Love the idea. Let's leave it out of this PR and keep it in mind for future work

PR #540, opened 2020-07-15
Move gatsby docs to guides

kendallstrautman:
This is a pretty big restructure. The gatsby docs are now entirely removed from the docs nav. 

I did my best to pseudo-translate the gatsby docs to individual guides. I didn't spend a ton of time digging through the actual content, just updated glaring things. I think there's room to trim down on these guides. Some of the content is great, but it could be more general and not necessarily Gatsby-specific.

**Next Steps** @jhuggett 

- Going through the actual content and see what we need / like — how can we trim the fat or condense / reorganize the information? 
- Test all those documented helpers, make sure they still work. 
- Convert the prose in the new guides to be actual 'guidespeak', some of them are very documentation style (see next point)
- Decide what information belongs in the package documentation or other docs instead of being a guide?
- Write gatsby packages docs 
- Add more of a disclaimer to the gatsby initial guide regarding running two dev servers, etc? Or decide where should this information go?

**_Note_: Style change**

I changed the guides titles on the [guides landing page](https://tinacms-site-next-msm4gefh1.vercel.app/guides) to be a plain paragraph tag instead of the H3 for readability.


kendallstrautman:
holy merge conflicts lol

PR #544, opened 2020-07-16
Adding delete action docs

logan-anderson:
Not sure where to put these docs so they are probably in the wrong spot. Just wanted to get this PR in so we could review the content and discuss the placement. 

EDIT: looking at #543 should this be added to the readme for react-tinacms-github


ncphillips:
I think I'll close this for now

PR #548, opened 2020-07-17
update nav to new structure

ncphillips:
This PR updates the navigation and it's styling. It also restructures (most) content accordingly. 

image


logan-anderson:
Looks really good! just noticed a couple of small things (not sure if they are fixable or not)

1. Is there no link to get back to the home page once you get into the docs
2. When you are clicking through the docs the nav always scrolls to the top
https://www.loom.com/share/904329b273ba4835969ee196e029bb1c

Besides those two things looks amazing!

spbyrne:
@logan-anderson For 1 we plan on adding 'Home' to the navbar when you're on docs pages.

For 2, I plan on moving search to below the 'Docs' header in the sidebar. When you scroll, the logo / search will be stuck to the top. That way, we can programmatically set the scroll position on pages where that's necessary without making the header hidden. This task is big enough to warrant a new PR.

PR #549, opened 2020-07-17
Nav change content org

ncphillips:
The base branch contains just the updates to the TOC

This contains the actual moving around of files to match that TOC


mittonface:
Some of the sidebar content gets pushed under other elements as you expand deeply. I.e., Expand Plugins->Fields and you'll see you lose the ability to click Content Creators, Screens and Toolbar widgets 

image

ncphillips:
Yeah that's kinda messed up. I think @spbyrne will take a look at correcting the navigation styling.

PR #563, opened 2020-07-23
Packages Page

jhuggett:
Adding package pages to the docs by pulling the READMEs from GitHub.


pages/packages/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+          previewError: { ...e }, //workaround since we cant return error as JSON
+        },
+      }
+    } else {
+      throw e
+    }
+  }
+}
+
+export default DocTemplate
```
ncphillips, line 22: That's super interesting

jhuggett: In a good way or a bad way?

utils/docs/getPackageProps.ts
```diff
@@ -0,0 +1,68 @@
...
+            return '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2)
+        })
+        .join('')
+    )
+}
+
+export async function getPackageProps({ preview, previewData }: any, slug: string) {
+
+    const filePath = path.join(process.cwd(), 'content/packages.json')
+    const file = JSON.parse(fs.readFileSync(filePath, 'utf8'))
```
ncphillips, line 25:
```suggestion
    const file = await readJsonFile('content/packages.json')
```

pages/packages/index.tsx
ncphillips: Good, I just didn't even think about re-using a template like that

utils/docs/getPackageProps.ts
jhuggett: It does, just had to export it

pages/packages/[slug].tsx
```diff
@@ -0,0 +1,96 @@
...
+import fs from 'fs'
+import { getPackageProps } from '../../utils/docs/getPackageProps'
+import { DocsNavToggle, DocsMobileTinaIcon, DocsContent } from 'pages/docs/[...slug]'
+
+  
+  export default function Packages(props) {
+    const frontmatter = {
+      title: 'packages',
+    }
+    const excerpt = 'A package for Tinacms.'
```
Comment (DirtyF) [inline pages/packages/[slug].tsx:27]: TinaCMS

kendallstrautman:
The layout and styles need to match the docs template. And I think we should add the title so it's consistent with the other docs...so we'll need to add the fm value in the READMEs. It would also be awesome to add the TOC if possible.

spbyrne:
I can handle the layout changes & adding the ToC. There's a lot of code to duplicate (I copied the new layout to guides) so I may refactor to remove that duplication for pages using docs layout/styling.

kendallstrautman:
Beat ya to the layout stuff - https://github.com/tinacms/tinacms.org/pull/570 lol but yeah let's do the TOC on another PR. I think you're right, there's some stuff we could extract 👌

I think all that's left is to have the title source from the frontmatter of the README file and remove those h1s on the READMEs

kendallstrautman:
Also @jhuggett do you have prettier set up? Just seeing some formatting weirdness

jhuggett:
@kendallstrautman I have prettier installed but I probably don't have it configured correctly

kendallstrautman:
lol I feel ya..... @spbyrne & @ncphillips can attest to my pains with prettier 🤦‍♀️

ncphillips:
This looks good to me. We can add more packages in separate PRs. Are we good to go here or are we waiting on a couple things?

kendallstrautman:
@ncphillips the only thing is sourcing the frontmatter title from the .md file, but we could also do that in another PR

kendallstrautman:
Oh and we should merge my layout adjustments

PR #569, opened 2020-07-23
guide: updates cms.enabled at auth step to check for undefined

kendallstrautman:
Since `props.pageProps.preview` is `undefined` until we set up the preview on the index page (at the next step), the CMS was automatically enabled and this was causing an issue where it wouldn't trigger the auth flow. This more strict check for `undefined` should make this step behave as described — allowing people to confirm their auth via cookies being set.

This change _should_ fix your issue @DirtyF, although it's slightly difficult to test since I didn't go through the guide again from step 1, so let me know.

### DIFF
--- a/content/guides/nextjs/github/configure-custom-app.md
+++ b/content/guides/nextjs/github/configure-custom-app.md
@@ -26,7 +26,10 @@ export default class Site extends App {
      * 1. Create the TinaCMS instance
      */
     this.cms = new TinaCMS({
-      enabled: props.pageProps.preview,
+      enabled:
+        typeof props.pageProps.preview === 'undefined'
+          ? false
+          : props.pageProps.preview,
       apis: {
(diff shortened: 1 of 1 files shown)


content/guides/nextjs/github/configure-custom-app.md
```diff
@@ -26,7 +26,10 @@ export default class Site extends App {
      * 1. Create the TinaCMS instance
      */
     this.cms = new TinaCMS({
-      enabled: props.pageProps.preview,
+      enabled:
+        typeof props.pageProps.preview === 'undefined'
+          ? false
+          : props.pageProps.preview,
```
ncphillips, line 32: This casts to `boolean`:
```suggestion
      enabled: !!props.pageProps.preview
```
`!!undefined` => `false`

PR #577, opened 2020-07-24
refactors toc & adds to packages pages

kendallstrautman:
For some reason, moving the toc file into `components/ui` broke some stuff — the active toc items weren't registering. So I reverted that change to get back to baseline. 

This PR creates a `toc_helpers` file in utils that extracts the 'active' toc item event listener logic into a higher-order function. This refactors guides and docs templates to use the function and also adds the toc to the packages page. 

On another note, I noticed one weird thing with the guides toc (not caused by these changes), but let's dig into that in another PR.


kendallstrautman:
Still a mystery to me @spbyrne, but it should work now

PR #582, opened 2020-07-27
fix: generate better slugs in docs ToC

### LINKED ISSUE #580: Docs table of contents: some anchors don't work
Some links in the table of contents don't link to the correct anchor, observed on the following page: https://tinacms.org/packages/next-tinacms-github

if you click on _getGithubPreviewProps_ in the table of contents, it links to `#_getgithubpreviewprops_` when it needs to link to `#getgithubpreviewprops`.

The fundamental problem is that our code to generate the anchors from headings and our code to generate the table of contents are completely separate.

We could try to strip markdown when generating slugs for the table of contents (the lib we use _mostly_ does this but stumbles on underscores,) but this feels like a flimsy solution

A better solution might be to either generate the ToC from the react-markdown AST if we can, or just wait until the markdown document is rendered and build the ToC by querying the DOM, thus avoiding generating slugs in two different places.

dwalkr:
fixes #580 

This is the "cheap" fix I suggested in the issue, but likely fits the 80+% use case as underscores seem to be the main problem

I plan in going through the docs front-to-back today, I'm gonna check all the ToC links before merging this

### DIFF
--- a/utils/docs/getPackageProps.ts
+++ b/utils/docs/getPackageProps.ts
@@ -2,6 +2,7 @@ import { getJsonPreviewProps, readJsonFile } from '../getJsonPreviewProps'
 import axios from 'axios'
 import toc from 'markdown-toc'
 const atob = require('atob')
+import { slugifyTocHeading } from './slugifyToc'
 
 const b64DecodeUnicode = (str: string) => {
   // Going backwards: from bytestream, to percent-encoding, to original string.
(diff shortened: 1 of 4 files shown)


dwalkr:
Already see one failure mode for this regex: it rejects sets of underscores with backslashes in between them, because `_this \_ is not emphasised text`, but will also reject `_this \# because of the backslash_`

PR #586, opened 2020-07-28
added lastModified to blog, docs, and guides

jhuggett:
Added last modify to content by using the mtime metadata on files. This may or may not work.

### DIFF
--- a/pages/blog/[slug].tsx
+++ b/pages/blog/[slug].tsx
@@ -21,7 +21,9 @@ import { getMarkdownPreviewProps } from 'utils/getMarkdownFile'
 import { InlineWysiwyg } from 'components/inline-wysiwyg'
 import { usePlugin, useCMS } from 'tinacms'
 import Toc from '../../components/toc'
-function BlogTemplate({ file, siteConfig, preview }) {
+import fs from 'fs'
+import path from 'path'
+function BlogTemplate({ lastModified, file, siteConfig, preview }) {
   // fallback workaround
   if (!file) {
     return <Error statusCode={404} />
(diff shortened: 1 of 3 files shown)


pages/docs/[...slug].tsx
```diff
@@ -96,6 +98,7 @@ function DocTemplate(props) {
               <DocGridContent ref={contentRef}>
                 <hr />
                 <InlineWysiwyg name="markdownBody">
+                  {props.lastModified ? `Last Modified: ${props.lastModified}` : ""}
```
Comment (ncphillips) [inline pages/docs/[...slug].tsx:101]: I think this works:
```suggestion
                  {props.lastModified && `Last Modified: ${props.lastModified}`}
```

pages/blog/[slug].tsx
```diff
@@ -116,11 +119,19 @@ export const getStaticProps: GetStaticProps = async function({
...
+
+    return {
+      props: {
+        lastModified,
+        ...previewProps.props,
+        siteConfig: { title: siteConfig.title },
+      },
+    }
+  } catch (error) {
+    throw error
```
Comment (ncphillips) [inline pages/blog/[slug].tsx:134]: Let's remove the `try {} catch () {}`

If the `catch` just throws the error then it's not really doing anything and it can just be thrown

pages/guides/[category]/[guide]/[step].tsx
```diff
@@ -211,16 +213,27 @@ export const getStaticProps: GetStaticProps = async function(ctx) {
...
+        preview,
+        currentGuide: guideMeta.data,
+        guideMeta,
+        markdownFile,
+        allGuides: await getGuideNavProps(),
+        tocItems,
+      },
+    }
+  } catch (error) {
+    throw error
```
Comment (ncphillips) [inline pages/guides/[category]/[guide]/[step].tsx:233]: Same here with the `try-catch`

PR #594, opened 2020-07-29
Adding Last Edited

jhuggett:
Adding the last edited date to docs etc.


pages/blog/[slug].tsx
```diff
@@ -38,6 +38,22 @@ function BlogTemplate({ file, siteConfig, preview }) {
...
+    const decorator = createDecorator(
+      {
+        field: /.*/,
+        updates: {
+          'frontmatter.last_edited': () => formatDate(Date.now())
+        }
+      }
+    )
+    return decorator(form.finalForm)
+  }, [form.id])
```
Comment (ncphillips) [inline pages/blog/[slug].tsx:54]: This is great! What if we made this a hook?

```tsx
useFormDecorator(form, {
  field: /.*/,
  updates: {
    'frontmatter.last_edited: () => formatDate(Date.now())
  }
})
```

PR #599, opened 2020-07-30
Typo on blog post

jhuggett:
Fixes a typo on the security post. This is really to see if the last edited piece is working as expected.


jhuggett:
> Let's see 👀
> 
> Note: I wonder why we don't store the last edited as a proper YAML date.

Didn't know there was a proper YAML date, I'll add that in with the refactoring I'm doing

DirtyF:
@kendallstrautman might want to style this date a bit better, It should not seem like it's part of the Questions? block [here](https://tinacms-site-next-git-blog-text-fix.tinacms.now.sh/blog/upgrade-notice-improved-github-security).

Capture d’écran 2020-07-30 à 17 23 28

PR #608, opened 2020-07-30
docs: updates field overview

kendallstrautman:
- Adds a list of default and external field plugins
- Adds more information on field definitions, the `name` and `component`
- Adds custom inline field example

I wrote these pretty fast. They could use a readthrough.

### DIFF
--- a/content/docs/plugins/fields.md
+++ b/content/docs/plugins/fields.md
@@ -131,7 +131,7 @@ The `name` property should be updated to reflect the different path:
 
 ### _component_
 
-The `component` field property can either be a string or the name of a field plugin, or a React Component.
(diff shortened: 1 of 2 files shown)


ncphillips:
We need a different phrase as "inline fields" now has two meanings lol

content/docs/plugins/fields.md
```diff
@@ -34,3 +34,150 @@ interface FieldConfig {
...
+
+### _name_
+
+The `name` property connects the field with the source data by providing the path to that content from the root of the source data.
+
+For example, say we had a JSON object:
+
+```json
+{
+  "headline": "Banana Pancakes"
```
ncphillips, line 92: Love me some Jack Johnson

PR #609, opened 2020-07-31
update focusRing references

spbyrne:
This adds a description of the focusRing API to all inline fields. This is tied to the [Focus Ring Fixes PR](https://github.com/tinacms/tinacms/pull/1358).


content/docs/ui/inline-editing/inline-image.md
```diff
@@ -40,14 +40,14 @@ There are two ways to use `InlineImage`, with and without children. If no childr
...
+| `uploadDir`  | Defines the upload directory for the image. All of the post data is passed in, `fileRelativePath` is most useful in defining the upload directory, but you can also statically define the upload directory.                                                                                                                        |
+| `previewSrc` | Defines the path for the src attribute on the image preview. If using gatsby-image, the path to the `childImageSharp.fluid.src` needs to be provided.                                                                                                                                                                              |
+| `focusRing`  | Either an object to style the focus ring or `false`, which hides the focus ring entirely. For styles, `offset` (in pixels) controls the distance from the ring to the edge of the group; `borderRadius`(in pixels) controls the [rounding](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius) edge of the focus ring. |
```
kendallstrautman, line 49: Note it could also be `true` with the default styles, right?

kendallstrautman: We should note what the default it

spbyrne: Ah, good call. I'll switch up this wording.

kendallstrautman:
We also need to check if anything needs changing on the[ inline blocks](https://tinacms.org/guides/general/inline-blocks/overview) guide — where it describes focus ring behavior

content/docs/ui/inline-editing/inline-image.md
spbyrne: Update to:

Either an object to style the focus ring or a boolean to show/hide the focus ring. Defaults to `true` which displays the focus ring with default styles. For style options, `offset` (in pixels) sets the distance from the ring to the edge of the component, and `borderRadius` (in pixels) controls the [rounded corners](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius) of the focus ring.

PR #612, opened 2020-07-31
Docs: Updates inline editing docs

kendallstrautman:
What this does:

- Reorganizes inline editing overview to show default inline fields first, then custom inline fields
- Updates Inline Blocks docs — they were out of date, especially the block template interface
- Added additional reading sections — Crosslinked to the Inline Blocks guide and other good stuff


content/docs/ui/inline-editing.md
```diff
@@ -20,96 +20,67 @@ _Inline Editing_ in Tina refers to editing values directly in the area they appe
...
 
-```tsx
+```diff
 import * as React from React
 import ReactMarkdown from 'react-markdown'
 import { useForm } from 'tinacms'
-import { Wysiwyg } from 'react-tinacms-editor'
-import { InlineForm, InlineField } from 'react-tinacms-inline'
++import { InlineForm, InlineTextField } from 'react-tinacms-inline'
++import { InlineWysiwyg } from 'react-tinacms-editor'
```
ncphillips, line 61: Since `InlineWysiwyg` isn't a default inline-field component I'm thinking we should remove it from the basic example

content/docs/ui/inline-editing.md
```diff
@@ -20,96 +20,67 @@ _Inline Editing_ in Tina refers to editing values directly in the area they appe
...
+          <ReactMarkdown source={modifiedValues.markdownContent} />
++       </InlineWysiwyg>
       </main>
     </InlineForm>
   )
 }
 ```
 
-## Using Pre-configured Inline Fields
+All Inline Fields expect a `name` prop, as with regular [Tina Fields](/docs/plugins/fields#name), the value should be the path to the data being edited by that field. Refer to individual inline field docs to see their specific properties.
```
ncphillips, line 79: Nice linking 👍

content/docs/ui/inline-editing.md
```diff
@@ -118,37 +89,29 @@ However, Tina provides a set of pre-configured Inline Fields that should **work
...
+```jsx
+<InlineForm form={formObject}>
+  <InlineField name="path-to-data">
+    {({ input, status }) => {
+      if (status === 'active') {
+        // we're in editing mode, show an editable interface
+      } else {
+        // we're not in editing mode, show the production layout
+      }
+    }}
```
ncphillips, line 109: There's no `status` on `InlineFieldRenderProps` anymore. 

Instead do `if (cms.enabled)`

kendallstrautman: oops. good catch

kendallstrautman:
should be good now 👌

PR #613, opened 2020-08-01
Content suggestion

laurenclark:
I think you've dispensed with the sidebar (for good?) so this part about opening one is misleading. :)

### DIFF
--- a/content/pages/home.json
+++ b/content/pages/home.json
@@ -38,7 +38,7 @@
       },
       {
         "_template": "setup_point",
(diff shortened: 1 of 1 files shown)


DirtyF:
Good catch. Tina now offers several ways to edit content, the sidebar is still a thing, but if you prefer to only use inline blocks, then you can click on the block you want to edit in the page directly. We're currently revamping the homepage (#576) we'll need to rephrase that part, thanks for your feedback.

PR #625, opened 2020-08-05
Link to final-form name docs

ncphillips:
We don't always explain home a field's name works. The final-form docs have a good description.

### DIFF
--- a/content/docs/plugins/fields.md
+++ b/content/docs/plugins/fields.md
@@ -2,8 +2,8 @@
 title: Field Plugins
 prev: /docs/plugins/forms
 next: /docs/plugins/content-creators
+last_edited: '2020-08-05T15:33:06.570Z'
 ---
-
 Fields are added to forms via the `fields` array and create the editing interface of a form.
 
 ## Field Config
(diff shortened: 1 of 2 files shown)


content/docs/plugins/fields.md
```diff
@@ -129,6 +129,8 @@ The `name` property should be updated to reflect the different path:
 }
 ```
 
+> Checkout [the final-form docs]() for a more detailed look at how the `name` field works.
```
DirtyF, line 132:
```suggestion
> Check out [the final-form docs](https://final-form.org/docs/final-form/field-names) for a more detailed look at how the `name` field works.
```

@ncphillips Is that the right link?

content/docs/ui/inline-editing.md
```diff
@@ -76,16 +76,18 @@ export function Page(props) {
 
 All Inline Fields expect a `name` prop, as with regular [Tina Fields](/docs/plugins/fields#name), the value should be the path to the data being edited by that field. Refer to individual inline field docs to see their specific properties.
 
+> Checkout [the final-form docs]() for a more detailed look at how the `name` field works.
```
DirtyF, line 79:
```suggestion
> Check out [the final-form docs](https://final-form.org/docs/final-form/field-names) for a more detailed look at how the `name` field works.
```

content/docs/plugins/fields.md
ncphillips: Yes. I forgot to hit save on the wysiwyg link ui

DirtyF: ENTER will also save the link.

PR #626, opened 2020-08-05
docs: adds guides to docs nav

kendallstrautman:


content/toc-doc.json
```diff
@@ -242,6 +242,73 @@
...
+        ]
+      },
+      {
+        "id": "/guides#gatsby",
+        "href": "/guides#gatsby",
+        "title": "Gatsby",
+        "items": [
+          {
+            "id": "/guides/gatsby/adding-tina/project-setup",
+            "href": "/guides/gatsby/adding-tina/project-setup",
```
ncphillips, line 300: These should maybe be slug instead of href

PR #628, opened 2020-08-05
Guide Index template updates

kendallstrautman:
- Converts the guides index to match the docs layout (without cards)
- Adds basic guides index md data (could use more descriptions)
- Replaces the guides nav with the docs nav
- Forces a scroll to guides section in docs nav when on the guides index  (if there's a better way to do this, let me know)
- Adds a toc to the index


spbyrne:
This is looking great, already a huge improvement. Nav scrolling is working for me, though we may tweak once we have the fixed nav header ready (working on that now).

kendallstrautman:
Cool, I think there will be conflicts with [this pr](https://github.com/tinacms/tinacms.org/pull/565) so maybe we should get that in first.

kendallstrautman:
Yeah I'm planning on digging into the index content on another PR 👌

PR #631, opened 2020-08-06
if a codeblock's tags contain "copy" then a copy button will be rendered

ncphillips:
If a fenced code-block has the "copy" tag then it will be copyable. It is this way becase not all code-blocks should be copy-able.

## Todo
- [x] needs styling @spbyrne 

## Examples

Non-copyable JS

    ```js
    const christmastPresent = new Banana()
    ```

Copyable JS

    ```js,copy
    const christmastPresent = new Banana()
    ```

Copyable Text

    ```copy
    Hello World
    ```

image


spbyrne:
code copy button

kendallstrautman:
we should add a note to our style guide or something showing how to do this

PR #652, opened 2020-08-11
Nav Scroll

spbyrne:
This takes the logic @kendallstrautman wrote for auto scrolling to the guides nav section and applies it to all nav items. If a nav link appears more than half way down the nav, the nav will load scrolled to that location when the user is on that page.

Screen Shot 2020-08-11 at 2 09 43 PM

### DIFF
--- a/components/DocumentationNavigation/NavSection.tsx
+++ b/components/DocumentationNavigation/NavSection.tsx
@@ -1,8 +1,7 @@
 import { useContext, useState, useMemo, useEffect, useRef } from 'react'
 import { useRouter } from 'next/router'
 import styled, { css } from 'styled-components'
-import { NavContext } from 'components/DocumentationNavigation'
-
+import { NavListContext } from './DocsNavigationList'
 import RightArrowSvg from '../../public/svg/right-arrow.svg'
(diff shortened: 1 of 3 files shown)


kendallstrautman:
Looks great! The only one I noticed it didn't work for was the [release notes](https://tinacms-site-next-qu6o4nusf.vercel.app/docs/releases)

components/DocumentationNavigation/NavSection.tsx
```diff
@@ -90,27 +89,33 @@ export const NavSection = (section: NavSectionProps) => {
 }
 
 const NavLink = ({ section, currentPage, router }) => {
-  const guidesTitleRef = useRef<HTMLElement>(null)
-  const sidebarNav = useContext(NavContext)
+  const linkTitleRef = useRef<HTMLElement>(null)
+  const sidebarNav = useContext(NavListContext)
 
   function scrollToGuides() {
```
kendallstrautman, line 95: Might wanna rename this function

spbyrne: Good call, updated.

spbyrne:
Looks like nav highlighting also doesn't work for release notes, likely related.

PR #653, opened 2020-08-11
Guide Index Updates

kendallstrautman:
I tried to give more context to the purpose of the guides, how they should be used and give better outlines of what information is in each guide. It is definitely a lot more to absorb but I think being descriptive in this index will help folks find the information they are looking for without digging into one individual guide at a time.


spbyrne:
I think the expanded intro & brief guide descriptions are great. The individual step links after the guide description look a bit cluttered, though. We don't have a ton of formatting options with markdown so I'm not sure how to improve that hierarchy.

ncphillips:
In the Django page they actually put "Step X" before each steps name. Maybe that would help?

image

kendallstrautman:
Agreed. Yeah I tried that and it looked crazy since there are multiple ones stacked on top. But I'll change it to see if it feels less weird than what it is now

kendallstrautman:
Screen Shot 2020-08-11 at 11 37 47 AM

spbyrne:
Yeah I think that works. Easier to decipher for sure.

kendallstrautman:
Screen Shot 2020-08-11 at 11 40 44 AM
 
if they aren't list items it could look like this

kendallstrautman:
I could also add breaks between the title / description and the steps but that feels pretty hectic 

Screen Shot 2020-08-11 at 11 42 59 AM

Now you can see why I was going nuts over formatting

kendallstrautman:
A look at a few in a row with the 'Step' added. 

Screen Shot 2020-08-11 at 11 47 56 AM

I'm torn, I like the steps cause you get to see more what is in each guide before actully navigating to it, but it does feel visually overhwelming. On the other hand, it gives a better sense of how long or short some of the guides are.

spbyrne:
So many options! Honestly any of those 3 would work.

ncphillips:
Yeah  those all would be fine. Kind of like the last one the most personally

spbyrne:
Here's an idea using a nested blockquote to custom style the step listt

Screen Shot 2020-08-12 at 10 34 31 AM

kendallstrautman:
Nice! @spbyrne I dig it

spbyrne:
Okay I applied that formatting to everything but it was a bit much so I tweaked the styling. I like it now, I think we're good to go.

Screen Shot 2020-08-12 at 4 01 05 PM

PR #672, opened 2020-08-17
upgrade tinacms to prerelease

ncphillips:


dwalkr:
will this fix https://github.com/tinacms/tinacms.org/issues/669 ?

ncphillips:
I think so yes

package.json
```diff
@@ -54,7 +54,7 @@
     "next": "^9.5.0",
     "next-seo": "^4.1.0",
     "next-svgr": "0.0.2",
-    "next-tinacms-github": "^0.27.0",
+    "next-tinacms-github": "^0.25.0-alpha.0",
```
DirtyF, line 57: https://github.com/tinacms/tinacms/blob/master/packages/next-tinacms-github/package.json#L3

ncphillips: Bad `yarn add` thanks

PR #676, opened 2020-08-17
Inline blocks copy code

ncphillips:


content/guides/general/inline-blocks/overview.md
```diff
@@ -18,7 +19,7 @@ To get started, you'll need to clone [the repo](https://github.com/tinacms/inlin
 
 At this point, it is set up with a TinaCMS instance and renders a simple `Hero` component with hard-coded data. All of the styles and images are configured as well, so when you add blocks they'll look nice 💅.
 
-```bash
+```bash,copy
```
dwalkr, line 22: I'm thinking maybe we should avoid making terminal commands copy-button supported as a rule-of-thumb? It technically works but is a little weird

ncphillips: What would be the reason? I've seen it before and I definitely appreciate it

DirtyF: Only the first line would be excecuted here

ncphillips: Works for me?

dwalkr: the main gotcha with multiline bash commands is that when you paste it in the terminal, it executes immediately instead of waiting for a user-inserted newline (hitting enter).

Also, the last line (doesn't contain a newline when copied) doesn't execute, so once the other commands finish, it's just sitting in the prompt waiting for the user to hit enter.

None of this is technically wrong, but it's a slightly clunky/odd experience IMO.

ncphillips: > the main gotcha with multiline bash commands is that when you paste it in the terminal, the copied newlines cause the preceding command to execute immediately instead of waiting for a user-inserted newline

hmm yeess this is a feature not a bug IMO but 🤷

dwalkr: Didn't say it was a bug, it's just how unix works.

Let me put it this way: I think a copy button on code blocks containing multiple terminal commands does more harm than good. Regardless, it is a very small issue. If you disagree strongly we can keep it.

ncphillips: Nah I'm good with that just gotta workup the energy to make the change lol

PR #683, opened 2020-08-18
Random Docs Updates

kendallstrautman:
## what this does

- removes duplicate sidebar / toolbar config options from cms overview
- makes fields optional in `useForm` config
- rephrases the ui introduction to better capture the other ui components involved in creating an editing interface (i.e. not just sidebar / toolbar)


content/docs/cms.md
```diff
@@ -93,13 +95,13 @@ export default function SomeComponent() {
 
 ## Disabling / Enabling the CMS
 
-The CMS can be enabled or disabled via methods or configuration on the CMS object.
+The CMS can be enabled or disabled via methods or configuration on the CMS object. Disabling the CMS prevents editing UI from rendering and forms from registering.
```
dwalkr, line 98:
```suggestion
The CMS can be enabled or disabled via methods or configuration on the CMS object. Disabling the CMS prevents the editing UI from rendering and forms from registering.
```

dwalkr:
I think making fields optional is a good idea, but it doesn't look like they currently are and I don't see a PR open to make that change. I think we should hold off on changing the docs until that is changed + released

content/docs/ui.md
```diff
@@ -9,9 +9,15 @@ consumes:
     details: Shows toolbar interface
 ---
 
-The `tinacms` package provides two possible user interfaces: the sidebar and the toolbar.
+There are two main pieces of UI that can construct or supplement the content editing interface: the sidebar and the toolbar.
```
dwalkr, line 12:
```suggestion
There are two main UI components that support the content editing interface: the sidebar and the toolbar.
```
just a suggestion to make this sentence a bit simpler

content/docs/plugins/forms.md
```diff
@@ -102,7 +102,7 @@ In addition to Final Form's options, Tina's form hooks accept the following addi
 interface FormOptions<S> {
   id: any
   label: string
-  fields: Field[]
+  fields?: Field[]
```
kendallstrautman, line 105:
```suggestion
  fields: Field[]
```

content/docs/plugins/forms.md
```diff
@@ -116,11 +116,11 @@ interface FormOptions<S> {
...
-| key | description |
-| --- | --- |
-| `id` | A unique identifier for the form. This should be derived from the content to distinguish it from other instances of the form. |
-| `label` | A label for the form that will appear in the sidebar. |
-| `fields` | An array of fields that will define the shape of the form and how content is edited. |
+| key                 | description                                                                                                                                   |
+| ------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
+| `id`                | A unique identifier for the form. This should be derived from the content to distinguish it from other instances of the form.                 |
+| `label`             | A label for the form that will appear in the sidebar.                                                                                         |
+| `fields`            | _Optional:_ An array of fields that will define the shape of the form and how content is edited.                                              |
```
kendallstrautman, line 123:
```suggestion
| `fields`            | An array of fields that will define the shape of the form and how content is edited.                                              |
```

PR #687, opened 2020-08-19
fix search results clipping bug

spbyrne:
Bug:
Screen Shot 2020-08-19 at 10 57 23 AM

Fix:
Screen Shot 2020-08-19 at 10 57 38 AM


DirtyF (approved):
Good catch, didn't spot that one!

Capture d’écran 2020-08-19 à 16 56 12

PR #691, opened 2020-08-19
Fix release notes creation

### LINKED ISSUE #686: Create release from releases page
When I browse https://tinacms.org/docs/releases I can't create add a new release, I have to browse to https://tinacms.org/blog to do it.

When I add a new release page, then I get redirected to a 404.

**Demo**: https://www.loom.com/share/5d2bb44f8666429791c83ec2fdb30686

dwalkr:
fixes #686

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -15,6 +15,8 @@ import {
 import { GlobalStyle } from 'components/styles/GlobalStyle'
 import 'components/styles/fontImports.css'
 import path from 'path'
+import { BlogPostCreatorPlugin } from '../tinacms/BlogPostCreator'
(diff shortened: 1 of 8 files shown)


pages/_app.tsx
```diff
@@ -42,7 +44,12 @@ const MainLayout = ({ Component, pageProps }) => {
     },
   }
 
-  const cms = React.useMemo(() => new TinaCMS(tinaConfig), [])
+  const cms = React.useMemo(() => {
+    const cms = new TinaCMS(tinaConfig)
```
ncphillips, line 48: why not just list them in the config?

dwalkr: that would be way too easy

PR #698, opened 2020-08-24
fix guides pagination

spbyrne:
The previous/next links weren't working for guides pages. It looks like the currentPath had a trailing slash but the individual step slugs didn't. This is a quick fix but probably not the best way to fix this. I'm not sure if the slug format changed or why this was previously working.


dwalkr:
@spbyrne Looks OK to me, I just moved the normalization to the place where the comparison is actually done, and did it to both sides.

PR #705, opened 2020-09-10
Fix doc code on useForm

jeffsee55:
`form` is omitted  from the return array but then used on the next line

 const [modifiedValues, form] = useForm(formConfig)

  usePlugin(form)


DirtyF (approved):
Good catch!

PR #715, opened 2020-09-22
chore: upgrade media to @next

kendallstrautman:


pages/_app.tsx
```diff
@@ -32,16 +29,15 @@ const MainLayout = ({ Component, pageProps }) => {
   const tinaConfig = {
     enabled: pageProps.preview,
     toolbar: pageProps.preview,
+    sidebar: pageProps.preview,
```
kendallstrautman, line 32:
```suggestion
```

kendallstrautman: We probably want to leave this the way it was...

dwalkr (approved):

PR #719, opened 2020-10-04
Rename prop

lukedigby:
The post prop should be referenced as initialPost, but in this particular code block it wasn't. This PR fixes this.


dwalkr (approved):

dwalkr:
Thanks!

PR #723, opened 2020-10-07
Update Adding Media Store Example

ncphillips:
The first example–adding the store to an existing CMS–as really simple.

The second example–adding the store when instantiating the CMS– was much more complex and had React hook code.

This PR simplifies the second example

### DIFF
--- a/content/docs/media.md
+++ b/content/docs/media.md
@@ -2,9 +2,8 @@
 title: Media
 prev: /docs/events
 next: /docs/apis
-last_edited: '2020-09-29T15:53:15.812Z'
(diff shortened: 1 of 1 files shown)


content/docs/media.md
```diff
@@ -2,9 +2,8 @@
 title: Media
 prev: /docs/events
 next: /docs/apis
-last_edited: '2020-09-29T15:53:15.812Z'
+last_edited: '2020-10-07T18:15:58.042Z'
```
ncphillips, line 5: @jhuggett I keep forgetting how awesome this is

DirtyF (approved):

PR #736, opened 2020-10-28
Add GithubMediaStore to imports

canrozanes:
Not having `GithubMediaStore` imported throws an error as its referred to further down


DirtyF:
Duplicate of #735 

Thanks for the edit.

PR #739, opened 2020-10-30
chore(deps): tinacms 0.32.1

DirtyF:
Upgrade to latest tina, to fix the sticky editor toolbar when editing a blog post:

https://www.loom.com/share/fb867436cd60411db533d70a1b3bd604


kendallstrautman (approved):

kendallstrautman (changes requested):
We also need to adjust the [`InlineWysiwyg` component](https://github.com/tinacms/tinacms.org/blob/master/components/inline-wysiwyg.tsx#L18) to pass a direct pixel value (e.g. 68px).

kendallstrautman (approved):

PR #769, opened 2020-12-16
FIX: typo in component name in code example

pixelmord:
No biggie, but tripped up the olde copy n' paster


DirtyF (approved):
Good catch! Thanks.

PR #772, opened 2021-01-02
Fix syntax error in inline editing example

connor-baer:
The example code on the [InlineGroup docs](https://tina.io/docs/ui/inline-editing/inline-group/) was [missing two commas](https://github.com/tinacms/tinacms.org/pull/772/files#diff-def6a26f6e603e530b7dfadc41350b7062e4cb3ca3971dbeb2e5e3a82d5a9104R43-R44).

I edited and committed through the TinaCMS interface on the website. Apparently, this caused other formatting changes to the code. Would you like me to revert those?


DirtyF (approved):
Thanks, for the edit @connor-baer

PR #798, opened 2021-02-22
docs: Add documentation on using next/image

### LINKED ISSUE #797: docs: next/image component usage
Add an example on how to use https://nextjs.org/docs/api-reference/next/image with Inline block Image

joeinnes:
Fix #797

### DIFF
--- a/content/docs/ui/inline-editing/inline-image.md
+++ b/content/docs/ui/inline-editing/inline-image.md
@@ -235,3 +235,60 @@ export function Hero({ data }) {
   )
 }
 ```
+
+### Using next/image
+
+Note that the same principles for _Gatsby Image_ can be applied to the _next/image_ component. This gives you the benefits of automatic image optimisation, but comes with some caveats. You can read more about how this works on [the Next.js docs](https://nextjs.org/docs/basic-features/image-optimization).
+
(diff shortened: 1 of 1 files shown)


content/docs/ui/inline-editing/inline-image.md
```diff
@@ -235,3 +235,60 @@ export function Hero({ data }) {
   )
 }
 ```
+
+### Using next/image
+
+Note that the same principles for _Gatsby Image_ can be applied to the _next/image_ component. This gives you the benefits of automatic image optimisation, but comes with some caveats. You can read more about how this works on [the Next.js docs](https://nextjs.org/docs/basic-features/image-optimization).
```
DirtyF, line 241: What caveats?

joeinnes: Primarily:

* you have to specify a width and height in pixels (or use `layout="fill"`, which [has its own drawbacks](https://github.com/vercel/next.js/discussions/18739#discussioncomment-121272))
* images are lazy-loaded by default (although this can be changed)
* you can't pass `style` to the `img` tag, in case you want to do something like add CSS filters, etc., you need to do this with classes and `className` 

I can expand here if it's better, but thought to link users to the Next docs instead - that way there's no need to make changes here if Vercel change something.

DirtyF: No need to expand, I just wanted to make sure what you're implying here. Thanks for making it explicit.

@tinacms/maintainers We'd use the [new inline-based ref](https://github.com/tinacms/tinacms/pull/1749) soon in those examples I guess.

dwalkr (approved):

PR #808, opened 2021-03-09
fix: Homepage SEO

DirtyF:
Provide a more descriptive Title and Description for people browsing off-context (SERP, Social Media, Slack, etc.)

note: This feels very nice to be able to edit SEO so easily with Tina ☺️

### DIFF
--- a/content/pages/home.json
+++ b/content/pages/home.json
@@ -1,6 +1,6 @@
 {
   "seo": {
-    "title": "Add real-time editing to your React site",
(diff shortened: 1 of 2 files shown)


spbyrne:
Maybe we should change the SEO format to put Tina before the page string so the browser tab title still makes sense. I'll swap that around and see if it looks good, and maybe drop 'CMS'.

spbyrne (approved):
👍

PR #821, opened 2021-03-31
fix: docs navigation

DirtyF:
- [x] Rename `Enterprise` to `Tina Cloud` in the docs main navigation
- [x] Redirect to [`/early-access`](https://tina.io/early-access/)


spbyrne (approved):
Nice, thanks 👍

PR #822, opened 2021-03-31
chore(deps): tinacms 0.39.0 + Next 10.1

DirtyF:
🐶 Dog Fooding latest version on our website. Works fine locally.


Enigmatical (approved):
Just did a quick rundown of the changes in [Next 10.1](https://nextjs.org/blog/next-10-1); seems all around a good update.  Nothing immediately concerning, but maybe more worth looking into?  Specifically, the changes to image optimization (but I believe we'll benefit from those inherently), Preview Mode might be useful (though our CMS has its own check for that; not sure if we're using Next Preview or not).

The only behavioral change is the Router auto-scrolling and that seems like the behavior you'd want.

👍

PR #853, opened 2021-05-05
Draft: Evolution of inline editing blog post

mitchmac:
Definitely needs some love and polish!

### DIFF
--- a/content/blog/evolution-of-inline-editing.md
+++ b/content/blog/evolution-of-inline-editing.md
@@ -5,11 +5,11 @@ author: Mitch MacKenzie & DJ Walker
 last_edited: '2021-05-05T10:00:00.000Z'
 ---
 
-Inline editing removes many usability hurdles that content editors face when they are working with traditionally decoupled CMS forms. Editors get to see their content changes immediately in the context of their actual website. But developers have been sharing the trade-offs associated with inline editing's current implementation with us.
+Inline editing removes many usability hurdles that content editors face when working with traditionally decoupled CMS forms. Editors get to see their content changes immediately in the context of their actual website. But developers have been sharing the trade-offs associated with inline editing's current implementation with us.
 
 We're continuing to learn from our development community and looking at better ways of integrating the superpowers that inline editing bestows.
 
-And let's not forget about sidebar editing in this discussion! The attractiveness of inline editing often overshadows the flexible set up and usability wins that we get with sidebar editing. **We recommend starting with [sidebar editing](/docs/getting-started/cms-set-up/#sidebar) to get up and running quickly.**
+And let's not forget about sidebar editing in this discussion! The attractiveness of inline editing often overshadows the flexible setup and usability wins that we get with sidebar editing. **We recommend starting with [sidebar editing](/docs/getting-started/cms-set-up/#sidebar) to get up and running quickly.**
 
 ## A review of the original approach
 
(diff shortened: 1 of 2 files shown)


mitchmac:
Preview of post here: https://tinacms-site-next-git-evolution-of-inline-editing-tinacms.vercel.app/blog/evolution-of-inline-editing/

Thoughts on using codesandboxes? They are a bit squished because of the content's container width but I like how the code is copy-pastable and hackable directly from the post.

DirtyF:
Those interactive code snippets are nice, I'm pretty sure we can style iframes to take the full width for instance if I add

```css
margin: 2em calc(50% - 50vw)
```

It gets displayed like this:

CleanShot 2021-05-05 at 14 41 41@2x

content/blog/evolution-of-inline-editing.md
```diff
@@ -0,0 +1,69 @@
...
+
+We can create editing-specific routes if we are concerned about the performance implications of TinaCMS being bundled with a website. Frameworks like [Next.js](https://nextjs.org) will optimize Javascript bundles so that TinaCMS is not included on routes that don't need it.
+
+The drawback of this approach is that we will need to duplicate aspects of our website's components between the normal view-only routes and the routes where TinaCMS is included.
+
+## What's next?
+
+We'll be making it clear in our documentation and guides that the original inline editing approach and Ref-Based inline editing are still experimental.
+
+Sidebar editing avoids most of the friction of integrating inline editing and provides many of the same usability wins. [Tina Cloud](/early-access/) auto-generates sidebar forms based on your content's schema and represents our current recommended best practices.
```
DirtyF, line 67: I would start by stating this positively in the introduction that we recommend sticking with sidebar editing for now in order to avoid any further breaking change, as we're still rethinking inline editing. 

We could link to the sidebar inline editing demo to point people to a working example.

content/blog/evolution-of-inline-editing.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Editing Routes
+
+We can create editing-specific routes if we are concerned about the performance implications of TinaCMS being bundled with a website. Frameworks like [Next.js](https://nextjs.org) will optimize Javascript bundles so that TinaCMS is not included on routes that don't need it.
+
+The drawback of this approach is that we will need to duplicate aspects of our website's components between the normal view-only routes and the routes where TinaCMS is included.
+
+## What's next?
+
+We'll be making it clear in our documentation and guides that the original inline editing approach and Ref-Based inline editing are still experimental.
```
DirtyF, line 65: What needs to be done here? Can we apply those documentation changes this week?

content/blog/evolution-of-inline-editing.md
```diff
@@ -0,0 +1,69 @@
...
+
+Sidebar editing avoids most of the friction of integrating inline editing and provides many of the same usability wins. [Tina Cloud](/early-access/) auto-generates sidebar forms based on your content's schema and represents our current recommended best practices.
+
+Stay tuned! (Or help out! TinaCMS is open-source and built with many hands. [Share an idea or solution](https://github.com/tinacms/tinacms/issues))
```
DirtyF, line 69:
```suggestion
Please do [share ideas, constraints or design patterns](https://github.com/tinacms/tinacms/issues) on how to bring a better visual editing experience to the Jamstack without sacrificing the codebase,  It's a discussion we need to have with the community.
```

content/blog/evolution-of-inline-editing.md
```diff
@@ -0,0 +1,69 @@
...
+We then call `useFieldRef` to acquire the ref that will be attached to our title element.
+
+<iframe src="https://codesandbox.io/embed/tina-ref-based-inline-editing-p8kx4?fontsize=14&hidenavigation=1&theme=dark&view=split&editorsize=65"
+     style="width:100%; height:500px; border:0; border-radius: 4px; overflow:hidden;"
+     title="tina-inline-editing"
+     allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
+     sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
+   ></iframe>
+
+### Editing Routes
```
DirtyF, line 57: I believe @jeffsee55 is using another pattern on the cloud starter, than only loads Tina after auth, is that worth mentioning?

spbyrne:
I've figured out a way to have wider iframes, but are there any other examples of iframes in the blog or in our docs to ensure t his won't break those cases?

content/blog/evolution-of-inline-editing.md
mitchmac: I think that's made possible by having only sidebar editing. I think it's a lot trickier with inline editing.

content/blog/evolution-of-inline-editing.md
mitchmac: @dwalkr @spbyrne I think we were going to try doing this on the cloud doc branch, were there any ideas for UI? I think a good first attempt could just be a prominent visual element placed at the top of each page's body

spbyrne:
Okay I made it wide for this case specifically so it won't break anything else. We can use the 'wide' class on iframes in other spots if we want, though.

content/blog/evolution-of-inline-editing.md
dwalkr: agreed on both counts! I've mentioned before that for Tina, marking something as experimental is not a downgrade so much as an implied guarantee that the other parts are more stable 😅

content/blog/evolution-of-inline-editing.md
mitchmac: @DirtyF Added a paragraph to the introduction

content/blog/evolution-of-inline-editing.md
```diff
@@ -0,0 +1,73 @@
...
+
+We can create editing-specific routes if we are concerned about the performance implications of TinaCMS being bundled with a website. Frameworks like [Next.js](https://nextjs.org) will optimize Javascript bundles so that TinaCMS is not included on routes that don't need it.
+
+The drawback of this approach is that we will need to duplicate aspects of our website's components between the normal view-only routes and the routes where TinaCMS is included.
+
+## What's next?
+
+We'll be making it clear in our documentation and guides that the original inline editing approach and Ref-Based inline editing are still experimental.
```
DirtyF, line 69:
```suggestion
We made it clear in our documentation and guides that the original inline editing approach and Ref-Based inline editing are still experimental and subject to further changes.
```

DirtyF (approved):

PR #861, opened 2021-05-12
Add tina cloud starter guide

jeffsee55:
This PR adds back the tina cloud starter, which is essentially just the broken-down version of the readme.

It also moves inline editing guide to an experimental section and hides the gatsby guides. Not sure if it's best to delete them altogether or just keep them around for now


DirtyF:
New images for the Netlify build settings
CleanShot 2021-05-13 at 12 40 14@2x

CleanShot 2021-05-13 at 12 39 53@2x

content/guides/tina-cloud/starter/local.md
```diff
@@ -0,0 +1,55 @@
...
+
+```
+yarn dev
+```
+
+> In your .env.local file `NEXT_PUBLIC_USE_LOCAL_CLIENT` should be set to `1`, other values can be ignored for now.
+
+This command starts the GraphQL server and the Next.js application in development mode. It also regenerates your schema types for TypeScript and GraphQL so changes to your `.tina` config are reflected immediately.
+
+One of the most interesting aspects of the Tina Cloud Content API is that it doesn't actually require anything from the Cloud to work locally. Since Tina is by default a Git-backed CMS, everything can be run from your local filesystem via the CLI. :sunglasses:
```
logan-anderson, line 38: : sunglasses : does not show up as an emoji on the web page (not sure why)

content/guides/tina-cloud/starter/structure.md
```diff
@@ -0,0 +1,113 @@
+---
```
logan-anderson, line 1: Is this page needed for the guide? Makes sense in the readme but not sure if it adds value to the guide.

Enigmatical (approved):
Looks good 😎

content/guides/tina-cloud/starter/local.md
DirtyF, line 38:
```suggestion
One of the most interesting aspects of the Tina Cloud Content API is that it doesn't actually require anything from the Cloud to work locally. Since Tina is by default a Git-backed CMS, everything can be run from your local filesystem via the CLI. 😎 
```

DirtyF (approved):

content/guides/tina-cloud/starter/structure.md
jeffsee55: I thought it would make sense to just have a bit of an explanation. I think we're aiming to have the same-ish set of instructions in both places

PR #875, opened 2021-05-25
Quick Start Docs Callout

spbyrne:
The text will need some tweaking. Maybe there's a way to do this with a shortcode instead of inline HTML?

Screen Shot 2021-05-25 at 5 09 53 PM

Preview: https://tinacms-site-next-git-quickstart-callout-tinacms.vercel.app/docs/


jeffsee55:
Oh that's muuuch better than what I had in mind...

content/docs/index.md
```diff
@@ -9,6 +9,15 @@ last_edited: '2020-11-26T14:37:54.481Z'
 - The **TinaCMS toolkit**, a suite of NPM packages that enable developers to quickly create a custom content management frontend for a website.
 - **Tina Cloud**, a hosted service designed to make Git-based content management more powerful and accessible for cross-functional teams.
 
+<div class="callout">
+<img className="learnImage" src="../img/tina-laptop.png" alt="" />
+  <div>
+  <h3>Ready to get started?</h3>
+  <p>Read our getting started guide to get set up in no time.</P>
```
DirtyF, line 16:
```suggestion
  <p>Get a Tina running website in no time!</p>
```

spbyrne: @DirtyF  How about "Get a website running with Tina Cloud in no time!"?

DirtyF: even better!

DirtyF:
I guess MDX would be the way to go if we want to use components in Markdown, but we don't support that do we?

content/docs/index.md
spbyrne: Done.

dwalkr:
Looks great to me. FYI this website does support shortcodes via `remark-shortcode` https://github.com/tinacms/tinacms.org/blob/master/README.md#shortcodes

however, they aren't supported by the WYSIWYG editor.

dwalkr (approved):

PR #885, opened 2021-05-31
Docs: call out Gatsby friction and confusing Gatsby references

mitchmac:


spbyrne (approved):
You may want another review but it looks good to me! 👍

DirtyF (approved):
Cool, thanks @mitchmac 
CleanShot 2021-05-31 at 21 27 23@2x

PR #899, opened 2021-06-03
fix: strip shortcodes from markdown when generating plaintext

dwalkr:


utils/blog_helpers.test.ts
```diff
@@ -0,0 +1,15 @@
...
+
+describe('stripMarkdown', () => {
+  describe('with shortcodes', () => {
+    const testDocument =
+      'Hello, {{ WarningCallout text="Tina Cloud is too cool for you" }}World!\n'
+    const expected = 'Hello, World!\n'
+
+    it('should remove shortcode markup', async () => {
+      const stripped = await stripMarkdown(testDocument)
+      console.log(expected, stripped)
```
Enigmatical, line 11: Lingering debug statement?

dwalkr:
```suggestion
```
good catch, thanks!

DirtyF (approved):

PR #901, opened 2021-06-08
Fixes and improvements for tina cli docs.

perkinsjr:
Fixed the example query to show the correct name of getPostDocument and adding where the data is available via localhost:4000.

### DIFF
--- a/content/docs/tina-cloud/cli.md
+++ b/content/docs/tina-cloud/cli.md
@@ -203,6 +203,8 @@ When you use Tina's GraphQL forms, we know about all of the relationships in you
 yarn run tina-gql server:start
 ```
 
+This will start the server on  `http://localhost:4001` and can be accessed using a GraphQL client on `http://localhost:4001/graphql`
+
 #### Query the content
 
 With a GraphQL client, make the following request:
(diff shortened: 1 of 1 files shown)


DirtyF (commented):
CleanShot 2021-06-08 at 18 10 13@2x

When running with current main branch the query that works for me is:

```
  getPostsDocument(relativePath: "voteForPedro.md"){
    data {
      __typename
      ... on Article_Doc_Data {
        title
        author {
          data {
            ... on Author_Doc_Data {
              name
              avatar
            }
          }
        }
      }
    }
```

perkinsjr:
Based upon the cli docs The following is available :

# global queries, these will be present regardless of the shape of your schema:
getDocument
getCollection
getCollections
# global mutations
addPendingDocument
updateDocument

# schema-specific queries.
getPostDocument
getPostList
getAuthorDocument
getAuthorList
# schema-specific mutations
updatePostDocument
updateAuthorDocument

Are available and if you follow the docs provided it does not contain it.

jeffsee55:
That's my bad, seems like I got mixed up between the plurality of `post`, having said that - I think ideally we educate users to call them by their singular name (ie. "post"), @perkinsjr @DirtyF does that make more sense? We can fix up whatever is needed to get it that way

DirtyF (approved):

DirtyF:
@jeffsee55 I'll update the CLI documentation, it will be less confusing if we keep those up to date with the current state of the starter, as we tell people to fork it.

I saw Altair GraphQL Client has a docs button that will show you the documentation, I wonder how we could make sure our docs stay up to date in a similar way.

PR #908, opened 2021-06-17
Fix: Broken link to loading states

AaronPowell96:
Fix a broken link in getting started tutorial to loading states, blocks, and markdown.
There is still one remaining broken link ```[Inline Editing](/guides/general/inline-blocks/overview)``` which I do not know where to point it to. Maybe: https://tina.io/guides/experimental/inline-blocks/overview/ ?


DirtyF (approved):
Thanks for the edit 🙏

PR #913, opened 2021-06-24
fix: update broken guide link

jsejcksn:
Ref: https://github.com/tinacms/tinacms.org/commit/374682d02cb307ebfc4baf8a3ab3b8d24bbfc4f3

Discovered by @samueldlay


samueldlay (approved):
This URL is the correct URL

DirtyF (approved):
Good catch, thanks for the edit, we marked inline editing as experimental, as it's likely to evolve before 1.0.

PR #934, opened 2021-07-09
fix links in sidebar so highlight will work

ashleyjanedoiron:
Sidebar used to highlight the current page but was broken. Updated the links in the sidebar to make it work again 🎉 

Fixes: #934 

Screen Shot 2021-07-09 at 2 42 45 PM


jeffsee55 (approved):
👍

PR #936, opened 2021-07-12
Remove 'Guides' from root navigation

ashleyjanedoiron:
Remove 'Guides' from site navigation. Will rework Docs section to include Guides subsection.


dwalkr (approved):
makes sense to me. note that guides are already included under the docs umbrella in their own distinct subsection, so this may be all that's needed

PR #946, opened 2021-07-21
Add docs on defineSchema anatomy

jeffsee55:


content/docs/schema.md
```diff
@@ -0,0 +1,289 @@
+---
+title: Anatomy of the Tina schema
+---
+
+The Tina schema is a way of defining the shape of your content for your entire site, with traditional content management systems you may have done this sort of content modeling via GUI. However given its tight coupling to Git, TinaCMS considers the filesystem the ultimate source of truth, and leverages a "content-modeling as code" approach. To that end, your schema is defined in a file called `.tina/schema.ts` (only `.ts` is supported for now).
```
Enigmatical, line 5:
```suggestion
The Tina schema is a way of defining the shape of your content for your entire site.  With traditional content management systems you may have done this sort of content modeling via GUI, however, given its tight coupling to Git, TinaCMS considers the filesystem the ultimate source of truth and leverages a "content-modeling as code" approach. 

To that end, your schema is defined in a file called `.tina/schema.ts` (only `.ts` is supported for now).
```

content/docs/tinacms-context.md
```diff
@@ -0,0 +1,182 @@
...
+          </TinaCMS>
+        }
+      >
+        <Component {...pageProps} />
+      </TinaEditProvider>
+    </>
+  )
+}
+
+export default Ap
```
Enigmatical, line 143:
```suggestion
export default App
```

Enigmatical (dismissed):
Just a few grammatical suggestions, but this looks great.  And it flows really well when talking about TinaCMS as a whole product (rather than delineating between TinaCMS and Gateway concepts). 👍

content/docs/cli.md
```diff
@@ -0,0 +1,132 @@
...
+
+Started Filesystem GraphQL server on port: 4001
+Visit the playground at http://localhost:4001/altair/
+Generating Tina config
+...
+```
+
+The below query can be run against your locally-running GraphQL server
+
+<iframe loading="lazy" src="http://localhost:3000/api/graphiql/?query=%7B%0A%20%20getDocument(collection%3A%20%22post%22%2C%20relativePath%3A%20%22voteForPedro.json%22)%20%7B%0A%20%20%20%20...on%20PostDocument%20%7B%0A%20%20%20%20%20%20data%20%7B%0A%20%20%20%20%20%20%20%20title%0A%20%20%20%20%20%20%20%20author%20%7B%0A%20%20%20%20%20%20%20%20%20%20...on%20AuthorDocument%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20data%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20name%0A%20%20%20%20%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D" width="800" height="400" />
```
DirtyF, line 124: Getting a connexion failed as I'm not running Tina locally, should be just do a screenshot here?
CleanShot 2021-07-26 at 10 25 04@2x

content/docs/cli.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+## Run the local GraphQL server
+
+Let's add some content so we can test out the GraphQL server
+
+#### Add an author
+
+```sh
+mkdir content && mkdir content/authors && touch content/authors/napolean.md
```
DirtyF, line 83:
```suggestion
mkdir content && mkdir content/authors && touch content/authors/napoleon.md
```

nit: we have this typo in our starter for a while now ([source](https://en.wikipedia.org/wiki/Napoleon_Dynamite))

DirtyF (approved):
LGTM

Enigmatical (dismissed):

logan-anderson (dismissed):

Enigmatical (dismissed):

logan-anderson:
@jeffsee55 Just noticed two bugs
1. The link to the toolkit in the overview page seems to be broken.
2. The overview text doesn’t highlight when link (not a big deal)
https://www.loom.com/share/c27f34c9d21c494fb8fc84ca59e36c8c

logan-anderson:
This is super minor and doesn’t have to be done but we maybe could but a CTA in the overview page (docs/tinacms-reference/) that links to a getting started guide if someone has not gone thought it before.

content/docs/schema.md
```diff
@@ -0,0 +1,291 @@
...
+
+export default defineSchema({
+  collections: [
+    {
+      label: 'Blog Posts',
+      name: 'post',
+      path: 'content/posts',
+      fields: [
+        {
+          type: 'text',
```
DirtyF, line 21:
```suggestion
          type: 'string',
```

shouldn't be that `string` now with the primitive types?

content/docs/schema.md
```diff
@@ -0,0 +1,291 @@
...
+      name: 'authors',
+      path: 'content/authors',
+      fields: [
+        {
+          type: 'text',
+          label: 'Name',
+          name: 'name',
+        },
+        {
+          type: 'text',
```
DirtyF, line 44:
```suggestion
          type: 'string',
```

content/docs/schema.md
```diff
@@ -0,0 +1,291 @@
...
+        },
+      ],
+    },
+    {
+      label: 'Authors',
+      name: 'authors',
+      path: 'content/authors',
+      fields: [
+        {
+          type: 'text',
```
DirtyF, line 39:
```suggestion
          type: 'string',
```

content/docs/schema.md
```diff
@@ -0,0 +1,291 @@
...
+        },
+      ],
+    },
+  ],
+})
+```
+
+# `collections`
+
+The top-level key in the schema is an array of _collections_, a `collection` informs the API about _where_ to save content. You can see from the example that a `posts` document would be stored in `content/posts`. You can supply either `fields` or `templates` to define shape of your collection. For the most part you'll likely just want to use `fields`
```
DirtyF, line 56: what does it mean for the most part, when would I want to use templates instead of fields?

logan-anderson: templates could be use when you have different types of the same thing. Kinda like [Polymorphism](https://en.wikipedia.org/wiki/Polymorphism_(computer_science)) If you are familiar with that. One example might be you have Feature pages, Content pages, and Blog pages but you want them all to be under the collection page. In this example all the different types of pages (content, feature, blog) would have different fields.

content/docs/media-cloudinary.md
```diff
@@ -0,0 +1,110 @@
+---
+title: Using Media with Cloudinary
+---
+
+Manage **Cloudinary media assets** in TinaCMS.
+
+## Installation
+
+```bash
+yarn add next-tinacms-cloudinary
```
logan-anderson, line 10:
```suggestion
yarn add next-tinacms-cloudinary @tinacms/auth
```

content/docs/media-cloudinary.md
```diff
@@ -0,0 +1,110 @@
...
+
+export const config = mediaHandlerConfig
+
+export default createMediaHandler({
+  cloud_name: process.env.NEXT_PUBLIC_CLOUDINARY_CLOUD_NAME,
+  api_key: process.env.NEXT_PUBLIC_CLOUDINARY_API_KEY,
+  api_secret: process.env.CLOUDINARY_API_SECRET,
+  authorized: async (req, _res) => {
+    try {
+      const user = await isAuthorized(req)
```
logan-anderson, line 85: Not super important but this will only work with Tina Cloud. Meaning if anyone working locally this will return unauthorized. As a fix you can add a check for the local_client env var and if they is true then return true.

logan-anderson (changes requested):

content/docs/schema.md
DirtyF: So if all your collection documents share the same fields, just use `fields`, i
if you need to use different templates inside a collection (e.g about page, pricing page, jobs page inside a `pages`collections) then use `templates`.

Seen no need to use fancy words to explain that 😉

content/docs/tinacms-context.md
```diff
@@ -0,0 +1,182 @@
...
+## `getStaticPropsForTina`
+
+TinaCMS is easiest to work with when you provide a predictable shape to the props you return from `getStaticProps`. `getStaticPropsForTina` enforces this shape for you:
+
+```tsx
+// pages/home.js
+import { getStaticPropsForTina } from 'tinacms'
+
+const getStaticProps = () => {
+  return getStaticPropsForTina({
```
logan-anderson, line 20: I thought we wanted this API to look something like this.

```ts

export const getStaticProps = () => {
  const tinaProps = getStaticPropsForTina({
    query: `
      query GetPostDocument($relativePath: String!) {
        getPostDocument(relativePath: $relativePath) {
          title
          body
        }
      }
    `,
    variables: {
      relativePath: 'hello-world.md',
    },
  })
return {
props: {
...tinaProps,
myOtherProp: "test",
},
}
```

logan-anderson (changes requested):
This is looking awesome!

content/docs/tinacms-context.md
jeffsee55: Thanks @logan-anderson, you're right that's the actual API

perkinsjr (approved):

perkinsjr (dismissed):
LGTM

logan-anderson (approved):
Looks good!

PR #955, opened 2021-07-27
Remove links to Vercel previews

jamespohalloran:
We were linking to old deployment previews. This caused me to take a bit of a detour and browse the wrong set of docs :)


DirtyF (approved):
oh!

PR #960, opened 2021-07-29
added cli docs

logan-anderson:
Added CLI and Tina init docs. I also changed "Anatomy of the Tina schema" to "content modelling" as I think its a bit easier to understand.


perkinsjr (approved):
LGTM

PR #961, opened 2021-07-29
Removing pieces from the navbar

perkinsjr:
Removing guides we don't need from the sidebar:

Removing TinaCMS totally and moving GitHub to experimental

### DIFF
--- a/content/toc-doc.json
+++ b/content/toc-doc.json
@@ -280,24 +280,6 @@
           }
         ]
       },
-
-      {
-        "id": "/guides#tinacms",
-        "href": "/guides#tinacms",
-        "title": "TinaCMS",
-        "items": [
-          {
-            "id": "/guides/tinacms/adding-tina/overview/",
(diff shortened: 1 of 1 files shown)


logan-anderson:
@perkinsjr maybe in the guides section instead of Tina Cloud, we could just call it TinaCMS?

perkinsjr:
@logan-anderson oops deleted the wrong name!

logan-anderson:
Maybe I am crazy but the names "Tina Starter" and "Getting started with TinaCMS" are very similar. If I was a new user I would be unsure which one to click. Maybe we would renames "Getting started with TinaCMS" to "Adding TinaCMS to an existing Next.js Project" or something along those lines. Thoughts @perkinsjr ?

perkinsjr:
Yeah I had the same feeling couldn't come up with a name that felt right, Just pushed a new one up "Adding TinaCMS to an existing Next.js app"

logan-anderson (approved):

PR #973, opened 2021-07-30
docs: API reference overview

DirtyF:
- Fix a 404
- Use text links instead of generic "learn more" links for better a11y.

### DIFF
--- a/content/docs/tinacms-reference.md
+++ b/content/docs/tinacms-reference.md
@@ -1,24 +1,25 @@
 ---
 title: Overview
-id: '/docs/tinacms-context'
+id: /docs/tinacms-context
+last_edited: '2021-07-30T16:08:34.541Z'
 ---
-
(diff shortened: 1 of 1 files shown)


content/docs/tinacms-reference.md
```diff
@@ -1,24 +1,25 @@
...
 title: Overview
-id: '/docs/tinacms-context'
+id: /docs/tinacms-context
+last_edited: '2021-07-30T16:08:34.541Z'
 ---
-
 ## Next.js API
 
-Enable live-editing on your Next.js site with TinaCMS. [Learn more](/docs/tinacms-context)
+Enable live editing on your Next.js site with TinaCMS. [Learn more](/docs/tinacms-context)
```
DirtyF, line 8: are we self looping here?

logan-anderson (approved):

PR #983, opened 2021-08-04
Guide fixes part 5

perkinsjr:
Adds some feedback.
new video
removes the old guide 
redirects to the new guide
add some call outs.


perkinsjr:
> Screen Shot 2021-08-04 at 10 27 53 AM
> 
> @perkinsjr Looks like this markdown is a little off here

Should be fixed I forgot the back ticks when reverting the change.

logan-anderson (approved):

PR #991, opened 2021-08-06
docs: Update headers on tina guides

jamespohalloran:
Update guide headings, as the starter guide name was misleading


spbyrne (approved):
Nice, good catch 👍

PR #995, opened 2021-08-09
Vercel integration documentation

jhuggett:
Adds a documentation page for our Vercel integration explaining what it's used for.

This is pretty bare-bones right now, as it a pretty simple use. Totally open to content suggestions and moving the page somewhere else. I'm not sure if we have a great spot to put this right now so I just added it to /pages as a one off.

And it isn't required to have a docs page for this but I think it's really nice to explain what we are using it for to the user.

Closes https://github.com/tinacms/teams-serverless/issues/735

### DIFF
--- a/pages/vercel-integration.tsx
+++ b/pages/vercel-integration.tsx
@@ -21,9 +21,7 @@ export default () => (
           <LastModified>Last Modified: August 9, 2021</LastModified>
           Our integration lets you easily deploy your sites to Vercel through the Tina dashboard.
           <br />
-          We will act on your behalf to create a project, configure environment variables, and trigger a deployment.
-          <br />
-          Once the project is created, feel free to revoke our integration's access, as it's only used for getting the project set up.
+          We will act on your behalf to create a project, configure environment variables, and trigger a deployment for the project we create.
         </Wrapper>
       </Section>
     </RichTextWrapper>
--- a/pages/vercel-integration.tsx
+++ b/pages/vercel-integration.tsx
@@ -1,0 +1,31 @@
+import React from 'react'
+import styled from 'styled-components'
+import {
+  Layout,
+  Hero,
+  Wrapper,
+  Section,
+  RichTextWrapper,
+} from 'components/layout'
+import { NextSeo } from 'next-seo'
+import { LastModified } from './terms-of-service'
+import Link from 'next/link'
+
+export default () => (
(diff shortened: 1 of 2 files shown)


pages/vercel-integration.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <Hero>Vercel Integration</Hero>
+    <RichTextWrapper>
+      <Section>
+        <Wrapper>
+          <LastModified>Last Modified: August 9, 2021</LastModified>
+          Our integration lets you easily deploy your sites to Vercel through the Tina dashboard.
+          <br />
+          We will act on your behalf to create a project, configure environment variables, and trigger a deployment.
+          <br />
+          Once the project is created, feel free to revoke our integration's access, as it's only used for getting the project set up.
```
DirtyF, line 26: but the next time you want to deploy a project you'll have to allow Tina too right?

jhuggett: Yeah, as it is right now, it's creating a new Vercel integration installation every time you go through the flow, so then revoking access has no negative effect. I am currently working on using a different auth method that's recommended, which likely will have some implications regarding this. But generally speaking a user wouldn't be deploying the starter multiple times so it's safe to say that they can remove access since we don't continually need it.

perkinsjr:
@jhuggett what is the status of this PR? is everything up to date?

jhuggett:
> @jhuggett what is the status of this PR? is everything up to date?

Just updated it now. It feels really light, but it is to the point about what we do with the integration. I think it's safe to merge and at least use it as a starting place. This isn't a necessary thing for us to have, but I think it is nice to have and having something is better than nothing.

perkinsjr (approved):

perkinsjr (changes requested):
Let's actually move this to 

docs/tina-cloud/vercel-integration 

this being on a single page doesn't make sense.

jhuggett:
> Let's actually move this to
> 
> docs/tina-cloud/vercel-integration
> 
> this being on a single page doesn't make sense.

Good point, moved it to `/docs/tina-cloud/vercel-integration` as suggested

jamespohalloran:
I'm going to close this for now as with all the docs updates (and Vercel path updates) it's gotten a bit stale. Feel free to reopen if you think it's still applicable @jhuggett

PR #1010, opened 2021-08-18
lazy load feature videos

spbyrne:
Trying to use the IntersectionObserver to lazy load videos [based on this guide.](https://web.dev/lazy-loading-video/)

I'm not 100% if it's working but it also doesn't appear to be broken, I would love another set of eyes on the code.

### DIFF
--- a/components/home/Features.tsx
+++ b/components/home/Features.tsx
@@ -310,9 +310,55 @@ const FEATURE_BLOCKS = {
 }
 
 export const FeatureVideo = ({ src }) => {
+  const videoRef = React.useRef(null)
+
+  React.useEffect(() => {
+    if (videoRef && videoRef.current) {
+      if ('IntersectionObserver' in window) {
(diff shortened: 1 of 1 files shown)


spbyrne:
On testing the preview deployment on my phone it appears like they all load in together still (any features w/ videos). Maybe the site loads and binds the intersection observer before the dom rendering catches up so it appears like all the videos are above the fold? Just a theory.

DirtyF:
Lighthouse says 14s to interact with the page but it doesn't seem to be related to the video
CleanShot 2021-08-18 at 19 03 33@2x

DirtyF:
On my Ipad the video is not above the fold but feels loaded when I scroll

logan-anderson:
It looks like it is not lazy loading but I definitely could be wrong: https://www.loom.com/share/692ccb0e32d84caf91cfc5d3366a3da2

PR #1019, opened 2021-08-20
fix: alpha to beta

DirtyF:
We're now in beta, our default URL has changed to app.tina.io

@jamespohalloran can we just call it Tina Cloud FAQ instead of Alpha FAQ?

### DIFF
--- a/content/docs/tina-cloud/alpha-faq.md
+++ b/content/docs/tina-cloud/alpha-faq.md
@@ -1,28 +1,27 @@
 ---
-title: Tina Cloud Alpha FAQ
+title: Tina Cloud FAQ
 ---
 
-## What is the Tina Cloud Alpha release?
+## What is the Tina Cloud Beta release?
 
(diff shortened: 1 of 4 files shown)


content/docs/tina-cloud/alpha-faq.md
```diff
@@ -45,8 +44,6 @@ The [Next.js starter](https://github.com/tinacms/tina-cloud-starter) can get you
 
 ## What is the pricing for Tina Cloud?
 
-There will be no cost for small teams to use Tina Cloud while it is in Alpha.
+Tina Cloud remains free until we go public. We are still working out our pricing model, and will offer a forever free plan for personal use.
```
jamespohalloran, line 47: @scottgallant is this line safe to say?

perkinsjr:
@DirtyF and @jamespohalloran instead of reusing the Alpha we have a new one https://tina.io/docs/tina-cloud/faq/ we should just redirect

perkinsjr (approved):

PR #1045, opened 2021-09-29
Make the full docs nav viewable when nested

jamespohalloran:
This still needs a lot of work, but wanted to get people's thoughts before I moved further. 

I find currently, it's easy to get lost in the docs, as there's not much context of where you are. 

I think if we can show the full nested context from the sidebar will help.

Notes:
- Guides pages currently blow up 
- We should probably do some considerable refactoring of the nav if we want to move forward. It's clear it's gone through quite a few iterations.

### DIFF
--- a/components/DocumentationNavigation/DocsLeftSidebar.tsx
+++ b/components/DocumentationNavigation/DocsLeftSidebar.tsx
@@ -28,8 +28,6 @@ export const DocsLeftSidebar = styled.div<{ open: boolean }>`
     background-attachment: local, scroll;
     background-repeat: no-repeat;
     background-size: 100% 1rem, 100% 1rem;
-    overflow-x: hidden;
-    overflow-y: auto;
     margin-right: -1px;
   }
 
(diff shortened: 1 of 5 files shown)


jeffsee55:
I like it, I would even wonder if we can take it further and just expand all of those  top-level items. It'd require a larger UI overhaul but something like GitBook docs seems more scannable ([example](https://docs.slatejs.org/))

spbyrne:
@jeffsee55 I'm working on a version that's more expanded, trying to balance that with something that's not just a huge list. I think having the final level only expand on navigation might help.

jeffsee55:
@spbyrne sounds good, there's also a lot that could potentially be deleted entirely - for example, we may want to remove about half of the items in the `toolkit` section

spbyrne:
Here's what it looks like so far: I have it auto-expanded for levels 1-3 and from there it expands if you're in that section (like fields). Feels pretty useable; I'll commit these changes so others can see it in action. 

Screen Shot 2021-10-05 at 5 07 36 PM

jamespohalloran:
The one thing that feels a bit weird to me with this is the quickstart guide nav link, which shows up twice in the nav. I imagine we'll soon be tweaking the getting started section soon anyways as we release app.tina.io/quickstart, so IMO the rest is a big enough improvement to ignore that for now.

PR #1048, opened 2021-10-14
Link to Quickstart Flow 

ashleyjanedoiron:
Link to Quickstart Flow from tina.io

### DIFF
--- a/content/pages/home.json
+++ b/content/pages/home.json
@@ -17,9 +17,9 @@
         },
         {
           "variant": "ghost",
-          "label": "Quickstart Guide",
+          "label": "Get Started",
           "icon": "",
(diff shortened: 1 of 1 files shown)


ashleyjanedoiron:
> The Github fix for the quickstart is merged to prod, so I think the last item we discussed waiting on is [tinacms/tina-cloud#966](https://github.com/tinacms/tina-cloud/issues/966)
> 
> I'd even argue we could probably point people at the quickstart before that last issue, (as this same behaviour exists regardless if your registering from Github through the quickstart, or dashboard) but up to you!

@jamespohalloran I see where Kelly has created a PR for that final issue so I'll just wait until it's merged to go live with this!

perkinsjr:
I am fine with this change if our goal is to have more people play with quickstart.

PR #1056, opened 2021-10-20
remove guides links

jamespohalloran:
Really cuirous of @perkinsjr and others' thoughts on this. 
I have a feeling that both guides ("using the tina starter" & "adding tinacms to an existing next.js app" ) should be covered by going through getting started > gaining some background info from basic features > going to production.
I think it will take some monitoring to see if that's actually the case, but if so, we should tweak the docs rather than leaning back on the guides. Right now, I find there are too many user paths to manage. I'm proposing as an experiment we remove links to them from our content, and remove them from the sidebar.

Thoughts?


perkinsjr:
I'd prefer if we kept the Adding Tina to Next. Versus removing it completely, we should have at least one guide that leans into how to use it Versus removing them. 

Happy to remove links from blogs etc, but keep it on the sidebar @jamespohalloran

jamespohalloran:
> I'd prefer if we kept the Adding Tina to Next. Versus removing it completely, we should have at least one guide that leans into how to use it Versus removing them.
> 
> Happy to remove links from blogs etc, but keep it on the sidebar @jamespohalloran

Yeah I'm good with that! I think us having removed the quickstart guide already makes things more manageable

PR #1087, opened 2021-11-17
Add version select

jamespohalloran:
Add a version select dropdown, so that users will be able to browse older site deployments.

This will allow us to have a cleaner split between the "legacy" docs, and the current promoted path on tina.io

TODO:
- [x] Use proper domain for our beta version
- [x] Get @spbyrne 's stamp of approval

Screen Shot 2021-11-17 at 12 21 22 PM

Screen Shot 2021-11-17 at 12 21 40 PM

### DIFF
--- a/components/DocumentationNavigation/DocumentationNavigation.tsx
+++ b/components/DocumentationNavigation/DocumentationNavigation.tsx
@@ -28,11 +28,13 @@ export function DocumentationNavigation({ navItems }: DocsNavProps) {
         onClick={() => setMobileNavIsOpen(!mobileNavIsOpen)}
       />
       <DocsLeftSidebar open={mobileNavIsOpen}>
-        <DocsSidebarHeader>
-          <DocsDesktopTinaIcon docs />
-          <VersionSelect />
+        <DocsSidebarHeaderWrapper>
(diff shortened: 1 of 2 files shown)


spbyrne:
Working on some styles for this :)

components/DocumentationNavigation/DocumentationNavigation.tsx
```diff
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import React, { useState } from 'react'
```
perkinsjr, line 1: Why are we importing React? We aren't using it anywhere as far as I can tell.

spbyrne (approved):
Works for me! When you switch docs versions it redirects you to the homepage, maybe it would be better UX to redirect to the docs index page but that's minor.

perkinsjr (changes requested):
Everything else looks great except importing React. Approved outside of that.

components/DocumentationNavigation/DocumentationNavigation.tsx
jamespohalloran: I think you're supposed to for any file with jsx

perkinsjr (approved):

PR #1099, opened 2021-11-22
Adding changes so the guide works:

perkinsjr:
- getPostsList is not a thing its getPostList
- Removing a # from a post
- Updating all queries and references


jamespohalloran (approved):
Good catches!

PR #1104, opened 2021-11-23
Make contextual-editing headings a bit more objective-based

jamespohalloran:


perkinsjr (approved):
Looks good! I wish our content extended a bit further so we don't have 2 lines for headings. But content is great.

PR #1106, opened 2021-11-24
Extending tina docs

jamespohalloran:
I felt that our "Extending Tina" section deserves to expand a bit, so I made it its own category with sub-docs around specific objectives:

Screen Shot 2021-11-24 at 3 42 15 PM

I think the docs could be improved around using formify or cmsCallback for non-field-customization ways, but I think we may need to first sort out how we want to structure our reference docs.

We also had a few reference docs, that felt a bit more like guides, so their content has been pulled up into this section


perkinsjr (dismissed):

perkinsjr:
Approved for when deploys don't fail.

perkinsjr (approved):

PR #1111, opened 2021-11-26
fix bad utils import

jamespohalloran:
fixed the broken indices creation from this PR: https://github.com/tinacms/tinacms.org/pull/1110/files

https://github.com/tinacms/tinacms.org/actions/runs/1509055606


perkinsjr (approved):
Approved

PR #1119, opened 2021-12-01
Updating SEO for Tina

perkinsjr:
Updating SEO for more realistic representation of Tina,


perkinsjr:
The rest are driven by pages content so we should be good.

PR #1120, opened 2021-12-01
Htmlproofer

jamespohalloran:
Add proofer to check for broken links on our site (and other things like missing image tags). 
Eventually, we likely will want to fail the build when htmlproofer fails, but there's a lot of issues to sort through first (which should probably sorted in a separate PR)

This action runs against the generated static export of the pages.

A few notes of weirdness. had to sort through here: 
- The "releases" docs & pages were also causing build issues, and weren't used so I remove them.
- "Static" pages (without getStaticProps) aren't added to exported out directory? So I added a dummy "getStaticProps" function to each
- I had to remove the `fallback: true` lines to get the static export to work on a few pages. I assume we had that for some Tina functionality. I checked, and creating new posts/docs isn't working on prod with Tina anyways, so I don't think removing the fallback's should affect existing behaviour.


perkinsjr (approved):
LGTM I am happy with the changes. 

We should investigate why page creation doesn't work though.

jamespohalloran:
> We should investigate why page creation doesn't work though.

I'm a bit torn on that. It's using a bit dated version of Tina with open-authoring, inline-editing, etc. Almost wondering if it would be better to wire up the Tina integration from scratch, and have it only available internally

PR #1122, opened 2021-12-01
Link to guide for tina init

jamespohalloran:
Right now, we're managing 2 paths within the intro doc, and it's a big hard to navigate where to send the user next. 
I think by sending users who already have a site to a specific guide, we can better segment users.


perkinsjr:
I'm fine with this change, it's a bit of an overlap with the next steps I was working on but I also had the same thoughts about why we were repeating the next steps for a site you already had.

perkinsjr (approved):

PR #1123, opened 2021-12-01
Update and Rework the Going to production section

kldavis4:
- Updates the Get Started link
- Updated a lot of the copy in this section
- Split the Dashboard doc into an intro and three separate pages
- Updates the screenshot with an updated version that has My Projects instead of My Apps
- Adds some detail to the projects page around Configuration / Overview tabs


perkinsjr (approved):
This looks great!

PR #1126, opened 2021-12-02
Create tina app doc

jamespohalloran:
- [ ] Once https://github.com/tinacms/tinacms/pull/2312#pullrequestreview-821962390 is deployed,
 we can add create-tina-app references to our site. 

I added a main CTA from the homepage

Screen Shot 2021-12-02 at 4 37 03 PM

@spbyrne may want to jump in and tweak the styles

### DIFF
--- a/components/home/Actions.tsx
+++ b/components/home/Actions.tsx
@@ -1,6 +1,15 @@
 import { WithCodeStyles } from 'components/layout/MarkdownContent'
 import React from 'react'
+import styled from 'styled-components'
 import { IconRight } from './Icons'
+
+const BashWrapper = styled.div`
+  code {
+    ::before {
+      content: '$ ';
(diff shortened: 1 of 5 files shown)


perkinsjr:
Yeah, I like this idea, but we should style it a bit more!

jamespohalloran:
I'm also starting to second guess if the user will have enough context of what to do once they enter the starter. Maybe it's best to instead link to a general "Getting Started" page which has the CLI command, the link to the quickstart, and a link to a tutorial for using Tina

jamespohalloran:
I'm going to close this PR for now actually, I think https://github.com/tinacms/tinacms.org/pull/1127 is probably a better incremental step, as the starters made through the CLI don't really have enough context of how to use Tina.

PR #1130, opened 2021-12-03
Update create-tina-app homepage CTA's

jamespohalloran:
Update homepage so that:
- Homepage has CLI CTA on main section
- Fix the "Command Line" section that shows both `create-tina-app` and `@tinacms/cli init`

Screen Shot 2021-12-06 at 12 58 31 PM

A couple things before we merge:

- [x] Could probably use some @spbyrne love
- [x] Give the user extra context from the CLI: https://github.com/tinacms/tinacms/issues/2321


spbyrne (approved):
Looks good to me! 👍

PR #617, opened 2020-08-04
Improve Nav Search

spbyrne:
- Adds working excerpt to search results
- Hides excerpts for title hits
- Reduce excerpt generation from 3000 to 200 characters
- Strip markdown from excerpts
- Improve hit components

Todo: fix issue with out of date index data (@dwalkr  is on it)

PR #1136, opened 2021-12-06
docs: index whole document, use snippets in search

### LINKED ISSUE #1044: Search only works for first 200 characters of each doc
On tina.io, when I search for something like `useDocumentCreatorPlugin`, there are no results. ~~It seems that we're not generating/updating the algolia index anymore~~

mittonface:
Fixes #1044 

Back in #617 a change was made so that we only indexed the first 200 characters of a doc. My best guess is that this was to make things less ugly when searching. 

This change switches things back to index the entire doc, but on the search side of things we use algolia's `<Snippet/>` functionality which I've configured to only display 50 words no matter how long the doc is.
Screen Shot 2021-12-06 at 11 43 24 AM

Let me know if 50 words is too many or if you remember there being some other reason for the change in #617

### DIFF
--- a/components/search/hitComps.tsx
+++ b/components/search/hitComps.tsx
@@ -1,9 +1,10 @@
-import React from 'react'
-import { Highlight } from 'react-instantsearch-dom'
-import { Hit } from 'react-instantsearch-core'
-import path from 'path'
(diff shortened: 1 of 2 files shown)


perkinsjr (approved):
This looks good to me.

PR #1138, opened 2021-12-07
Pull vercel env for our branch config

jamespohalloran:
We should try and use Vercel's system environment variables if available to populate the TinaCMS config. Most people will set Tina up on a branch initially rather than main, so this allows the user to forget about manually setting/switching branches in code

### DIFF
--- a/content/docs/tina-cloud/connecting-site.md
+++ b/content/docs/tina-cloud/connecting-site.md
@@ -7,15 +7,15 @@ Once you've created a project within the **Tina Cloud**, the next step is to con
 
 ## Enabling Tina Cloud in TinaCMS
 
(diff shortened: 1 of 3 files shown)


perkinsjr (approved):
LGTM

jeffsee55:
Yeah I agree, I was wondering if we should add this to the starter 🤔 . Obviously that breaks down immediately when not using Vercel

jamespohalloran:
> Yeah I agree, I was wondering if we should add this to the starter 🤔 . Obviously that breaks down immediately when not using Vercel

I think if we can show an error for when branch is not set (related to: https://github.com/tinacms/tinacms/issues/2329), it would reduce friction, (instead of using "main" by default). That way, I think it's easy enough to replace the Vercel environment variable once they know what's happening.

PR #1139, opened 2021-12-08
Tina MDX blog post

perkinsjr:
This will need to be merged on 12/09

### DIFF
--- a/content/blog/using-the-power-of-mdx-with-tina.md
+++ b/content/blog/using-the-power-of-mdx-with-tina.md
@@ -1,0 +1,227 @@
+---
+title: Using the power of MDX with Tina
+date: '2021-12-09T12:09:32-04:00'
+author: James Perkins
+last_edited: '2021-12-09T12:09:32-04:00'
+---
+# Using the power of MDX with Tina
+
+Tina allows content teams and developers to work at a fast pace and removes the friction between static sites and editing content. We took this approach a step further with the release of our MDX support, which allows developers to create reusable components and content teams to use them whenever they need to. This blog post will show you how to add Tina to your site, then how to create and use components to Tina.
+
+### Project setup.
+
+Step 1: Create a project with tailwind
+``` bash,copy
+npx create-next-app -e with-tailwindcss tina-demo
+cd tina-demo
+```
+Step 2: Add Tina to the project
+
+Use the following command inside of the project to add all the Tina dependencies and wrap the application, ready to be used.
+
+``` bash,copy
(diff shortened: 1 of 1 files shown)


content/blog/using-the-power-of-mdx-with-tina.md
```diff
@@ -0,0 +1,220 @@
...
+``` bash,copy
+npx create-next-app -e with-tailwindcss tina-demo
+cd tina-demo
+```
+Step 2: Add Tina to the project
+
+Use the following command inside of the project to add all the Tina dependencies and wrap the application, ready to be used.
+
+``` bash,copy
+npx @tinacms/cli@latest init
```
jeffsee55, line 23: Should they choose to override the _app.js when this command asks them?

content/blog/using-the-power-of-mdx-with-tina.md
```diff
@@ -0,0 +1,220 @@
...
+```
+Finally, we can update your `TinaMarkdown` component to pass the components for our users to use.
+
+```javascript,copy
+const components = {
+    Callout: (props) => {
+      return <Callout callout={props} />;
+    }};
+```
+
```
jeffsee55, line 213: This should be `<TinaMarkdown content={content} components={components} />`

jeffsee55 (commented):
My only qualm with this is that we're showing them how to add this to an existing site (with `tinacms init`) but then we don't actually show them anything related to _their_ existing site's code since this is all done for the page we generated for them. So why not just use `npx create-tina-app` and use the barebones starter give them less things to think about?

content/blog/using-the-power-of-mdx-with-tina.md
perkinsjr: Good catch! I had to redo the code because docs ended up breaking all the code off.

perkinsjr:
@jeffsee55 
The reason we aren't doing tina-create-app is that I want to write about how we have now made it easier to create applications. This then allows us to go back and update this.

content/blog/using-the-power-of-mdx-with-tina.md
perkinsjr: Yup. I should add that.

jeffsee55:
Do you mean as a separate blog post? Would it make sense to drop that in before this one?

perkinsjr:
> 

 Yup, a completely separate post. and no people tend not to read before December - January so this is under the radar. With that post going out early next year.

jeffsee55 (approved):

PR #1142, opened 2021-12-13
Add Tina Playground for references

jeffsee55:


jeffsee55:
@jamespohalloran I think this is safe to merge but there are still a few things I want to get done. One thing to watch out for is the playground is pointing to _this_ vercel deployment on this branch. And even when this merges it still will, but something we'll want to update when we can.

Remaining
- Mobile behavior, I think we'll show the tabs and not allow editing, preview _should_ work.
- Missing some explanations on formatting for things like dates
- Linking back to the content modeling section, and also cross-linking to the tina core field plugins so show which ones are used for each shape of your schema.

perkinsjr (approved):

jamespohalloran (approved):

PR #1156, opened 2022-01-04
Use apiURl prop

jeffsee55:
This API definitely feels better, but it's still a bit awkward when using the environment's branch. Did my best to offer some docs [on that](https://tinacms-site-next-git-use-api-url-tinacms.vercel.app/docs/tina-cloud/connecting-site/#using-the-deployment-branch), any insight welcome!

Goes with this PR https://github.com/tinacms/tinacms/pull/2377

EDIT: I've updated things so that the client id isn't stored as an env var. Maybe that helps a bit.

cc @jamespohalloran

### DIFF
--- a/content/docs/tina-cloud/connecting-site.md
+++ b/content/docs/tina-cloud/connecting-site.md
@@ -3,35 +3,63 @@ title: Connecting the site
 next: '/docs/tina-cloud/faq'
 ---
 
-Once you've created a project within the **Tina Cloud**, the next step is to connect your site. Once connected, your project's editors will be able to save content directly to it's GitHub repository, entirely from within your site.
+Once you've created a project within the **Tina Cloud**, the next step is to connect your site. Once connected, your project's editors will be able to save content directly to its GitHub repository, entirely from within your site.
 
 ## Enabling Tina Cloud in TinaCMS
 
 In the [Contextual Editing doc](/docs/tinacms-context/), we showed you how the Tina context is setup on your site.
 
-To now enable Tina Cloud on your site, we just need to tweak a few properties to make the site editable:
+To use Tina Cloud, change the `apiURL` to:
 
-| property        | description                                                                                                                                                                           |
-| --------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `clientId`      | This identifier associates your site with a Tina Cloud project. You can find your clientId on the [Dashboard](https://app.tina.io/projects) by viewing your project's configuration.  |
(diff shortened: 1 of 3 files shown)


jamespohalloran:
> EDIT: I've updated things so that the client id isn't stored as an env var. Maybe that helps a bit.

I wonder if users may miss that clientID as is. 

> This API definitely feels better, but it's still a bit awkward when using the environment's branch. Did my best to offer some docs on that, any insight welcome!

I could go either way on this, but could export a helper to generate that prod URL:

```
import { formatProdApiUrl } from tinacms

const apiURL = process.env.NODE_ENV
    ?  formatProdApiUrl({ branch: process.env.NEXT_PUBLIC_VERCEL_GIT_COMMIT_REF, clientID: 'YOUR-CLOUD-URL-HERE' })
    : 'http://localhost:4001/graphql'
```

jeffsee55:
A helper is interesting, still - I like the simplicity of "this URL or that URL". 

> I wonder if users may miss that clientID as is.

Maybe we should just pull out the clientID into a variable for clarity:

```ts
const branch = process.env.NEXT_PUBLIC_VERCEL_GIT_COMMIT_REF
const clientId = "YOUR-CLIENT-ID-HERE"
const apiURL = branch
    ? `https://content.tinajs.io/content/${clientId}/github/${branch}`
    : 'http://localhost:4001/graphql'
```

jamespohalloran:
Yeah I think that's clearer! 

> A helper is interesting, still - I like the simplicity of "this URL or that URL".

yeah agreed. We can always consider adding the helper down the road if there are issues.

content/docs/tina-cloud/connecting-site.md
```diff
@@ -3,35 +3,61 @@ title: Connecting the site
...
+`<myClientId>` is the value from the Tina Cloud dashboard, and `<myBranch>` is the branch which you wish to communicate with.
 
-### Example
+We recommend storing this URL in an environment variable so you can switch between `localhost` and Tina Cloud easily. Note that it's safe for this variable to be public so if you're using Next, prefix it with `NEXT_PUBLIC_` for client-side access.
 
 ```tsx
 // pages/_app.js
 import TinaCMS from 'tinacms'
 
+const apiURL = process.env.NEXT_PUBLIC_TINA_API_URL
```
jamespohalloran, line 26: Here you recommend using an environment variable, but later you recommend doing it inline with a dynamic branch. 

I think this may be simpler if we leave out the environment variable section (I think devs will identify on their own if they want to use environment variables).

jamespohalloran (approved):
Approved for when the tinacms package gets published!

jamespohalloran (approved):

PR #1191, opened 2022-01-25
update media docs

logan-anderson:
Noticed a small issue with the media docs. The way we where documenting it would make larger bundle sizes (on production routes).


jamespohalloran (approved):
Good catch!

PR #1195, opened 2022-01-27
update edit state docs

logan-anderson:
Just noticed a small issue with the docs that might confuse some folks.

### DIFF
--- a/content/docs/tinacms-context.md
+++ b/content/docs/tinacms-context.md
@@ -86,28 +86,25 @@ export default TinaAdmin;
 
 ### Manually toggling via `useEditState`
 
-You can manually enter and exit edit mode by tapping into the `useEditState` hook. A common pattern is to place this hook on an "admin" page, which simply puts you into edit mode and sends you back to the page you were on.
+You can manually enter and exit edit mode by tapping into the `useEditState` hook. You should **always** have a  `/pages/admin/[[...tina]].{js,tsx}` file but in some cases you may want to go into edit mode without going to the `/admin` or `/admin/logout` routes. 
+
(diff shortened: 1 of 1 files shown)


jamespohalloran:
One thing I was wondering is if people still often use this useEditState hook now that we provide a bit more out of our admin. In this [PR](https://github.com/tinacms/tinacms.org/pull/1186/files#diff-7349b5c7f0dd200645c64944973bc6ab5784a61c7ba52e4e212725a44579390dL87), I was thinking about removing it altogether from this doc (and maybe moving it to a reference doc if people request it).

What's your thoughts @logan-anderson, do you think it's still relevant?

logan-anderson:
> One thing I was wondering is if people still often use this useEditState hook now that we provide a bit more out of our admin. In this [PR](https://github.com/tinacms/tinacms.org/pull/1186/files#diff-7349b5c7f0dd200645c64944973bc6ab5784a61c7ba52e4e212725a44579390dL87), I was thinking about removing it altogether from this doc (and maybe moving it to a reference doc if people request it).
> 
> What's your thoughts @logan-anderson, do you think it's still relevant?

We have gotten asked a couple questions of folks wanting to add "custom logic" when in edit mode I think it is still relevant. What do you think adding it to the reference section?

logan-anderson:
> Yeah I think either reference, or in the "extending tina" section might make sense. I'm also a bit unsure of when someone would want to manually use that hook though, so maybe adding context around a real-world example might help.

Yeah definitely! 

> Also, does the user need to be logged in first for it to work?

Nope! this will automatically login the user with the auth model (if they don't successfully login it will set the edit state back to false)

PR #1196, opened 2022-01-27
add section on rendering markdown to markdown field docs

### LINKED ISSUE #1173: Document rendering markdown
Our [MDX docs](https://tina.io/docs/editing/mdx/) show an example of rendering out the rich text within a page.

Our [Markdown docs](https://tina.io/docs/editing/markdown/) should do something similar (likely using react-markdown)

dwalkr:
closes #1173


logan-anderson:
@dwalkr This is awesome and very clear. Would it make sense to have somewhere on our docs (could be apart of the PR but it does not have to be) to explain the difference between the rich text and the markdown editor? (Raw markdown string vs using an AST from `type: "rich-text"`)

PR #1222, opened 2022-02-02
upgrade next

jamespohalloran:
Upgrade NextJS, so we can use Node >12 (which should fix https://github.com/tinacms/tinacms.org/pull/1221)
There were some code-splitting breakages with the NextJS upgrade, so I had to tweak a few imports (and start building our rss feed differently)


ashleyjanedoiron (approved):
LGTM!

PR #1223, opened 2022-02-03
Add useTIna in next guide

jamespohalloran:
Update our guide to use useTina.

I think this could be made more clear, but I think we may want to revisit this whole guide in the near future anyways, as enough has changed since it was initially written.

### DIFF
--- a/content/guides/tina-cloud/add-tinacms-to-existing-site/next-changes.md
+++ b/content/guides/tina-cloud/add-tinacms-to-existing-site/next-changes.md
@@ -17,14 +17,14 @@ So based upon the `getPostList` we will want to query the `sys` which is the fil
 
 ```graphql,copy
 query {
-  getPostList{
-    edges{
-      node{
-        sys{
-           basename
-          }
+  getPostList {
+    edges {
+      node {
+        sys {
+          basename
         }
       }
+    }
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


content/guides/tina-cloud/add-tinacms-to-existing-site/next-changes.md
```diff
@@ -345,6 +345,34 @@ export default function Post({ data, slug}) {
...
 
+### Adding contextual editing:
+
+At this point, your page is rendering with the content loaded from Tina. We now want to setup Tina's contextual editing so that you can edit the content via the sidebar.
+
+Let's update your page component like so:
+
+```diff
+- export default function Post({ data, slug}) {
++ export default function Post({ data: initialData, slug, query}) {
```
jeffsee55, line 356: Aren't we suggesting that they no longer need to pass `query` (or `variables) through here? Could also then remove those from the `getStaticProps` return in the code snippet above

jamespohalloran: Yeah technically. I kind of just found I was running into a ton of friction on this page, that I kind of just patched this quick to get it working. I'm going to make a TODO to revisit this guide soon.

logan-anderson (approved):

PR #1225, opened 2022-02-04
Edit home page with tina

logan-anderson:
First go at this will just use blocks but we can always switch to MDX later. 

~~NOTE: this is only working in "local" edit mode. Someone will have to setup a tina cloud project for this to use.~~

It will work in prod editing once it is merged. (Thanks @jamespohalloran )


spbyrne:
I'm working on that sidebar header overlap issue, but in the meantime I've figured out we can fix it by providing a label in the schema, which I've pushed up.

logan-anderson:
Thanks @spbyrne I just added tina cloud as well so it should be good to go when we merge into main.

logan-anderson:
@spbyrne Looks like there is an issue with the command line QuickStart part of the home page. (It might be an issue with the generated client. Let  me look into it)

Screen Shot 2022-02-04 at 10 29 54 AM

jamespohalloran (approved):
Nice!

PR #1230, opened 2022-02-07
Add docs on querying blocks

### LINKED ISSUE #1171: Add context around queries to our "Editing with blocks" docs
Currently our Blocks editing section has half the information. Based upon what is provided you have no idea what the query should be. This leads users down a bad path. We should provide the query, otherwise the code snippets provided make no sense. Based on the code the query could look like this:

```
getPageDocument(relativePath: "path.md"){
        data{
          blocks{
            __typename
          }
        }
        }
}
}
```

However you need a fragment for each block. which isn't easily found because our example application that this is pulled from uses the client sdk.

Personally, I spent 30-40 minutes figuring out that I needed fragments for each block to make it work correctly.

-- TODO:

- [ ] Add missing query documentation
- [ ] Add a playground example

dwalkr:
closes #1171

### DIFF
--- a/content/docs/editing/blocks.md
+++ b/content/docs/editing/blocks.md
@@ -158,6 +158,8 @@ query getPagesDocument($relativePath: String!) {
 }
 ```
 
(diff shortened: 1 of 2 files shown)


content/docs/editing/blocks.md
```diff
@@ -125,6 +125,43 @@ We have defined the structure of our 3 blocks (`content`, `features`, `hero`), a
 
 > Note, since our contentBlock uses the [markdown plugin](/docs/reference/toolkit/fields/markdown/) which is not registered out of the box, we will need to manually import the `react-tinacms-editor` plugin.
```
logan-anderson, line 126: I know its not part of this PR but you could probably link to this section (guides/tina-cloud/add-tinacms-to-existing-site/markdown-plugin/) of the guide to provide more context of how to add the `react-tinacms-editor` plugin.

content/docs/editing/blocks.md
```diff
@@ -125,6 +125,43 @@ We have defined the structure of our 3 blocks (`content`, `features`, `hero`), a
...
 
+## Querying Block Data
+
+Because each item in a list of blocks can have a unique schema, querying this data isn't as straightforward as other types of fields. We'll need use GraphQL's fragment syntax to query the appropriate data shape for each block type.
+
+The fragment names are automatically generated based on the collection name and parent field name. For example, if the collection is `pages`, the field is `blocks`, and the block's name is `hero`, the fragment will be named `PagesBlocksHero`.
```
logan-anderson, line 132: Maybe add something that points to the fact that this maps to querying `__typename`

logan-anderson: nvm its at the bottom

logan-anderson (approved):
Looks good to me. I left one comment that you can add if you want to but does not need to block this from going out.

PR #1262, opened 2022-03-16
Migrate blog content management to Tina Cloud

dwalkr:
The MDX parser is not quite as resilient as Remark, and some nested inline elements that were OK with Remark now break the MDX parser. Seems to mostly concern anchor tags nested inside of strong/emphasis elements. The arbitrary formatting changes in the PR exist to work around these parsing issues.


spbyrne:
I got a strange error when navigating to edit a post from admin, so I adding route mapping to take you to the post directly to get around the issue.

spbyrne:
Screen Shot 2022-03-18 at 12 15 25 PM

For the 'previous'/'next' post reference fields, the 'edit in CMS' links take you to the admin edit page, which presents the same error I encountered before implementing route mapping.

spbyrne (approved):
This is awesome. I fixed the only issues I came across, otherwise seems good to go!

logan-anderson (approved):
Looks good to me as well.

perkinsjr (approved):
LGTM 🚀

PR #1263, opened 2022-03-16
Cms to contextual

perkinsjr:
CMS To contextual editing blog post.

### DIFF
--- a/content/blog/from-cms-to-contextual.md
+++ b/content/blog/from-cms-to-contextual.md
@@ -135,7 +135,7 @@ export async function getStaticPaths() {
 
 Remove all the code inside of this function and we can update it to use our own code. The first step is to add an import to the top of the file to be able interact with our graphql, and remove the `getPostBySlug` and `getAllPosts` imports we won’t be using:
 
-```bash
+```diff
 //other imports
 .....
 - import { getPostBySlug, getAllPosts } from '../../lib/api'
@@ -327,6 +327,7 @@ import {useTina} from 'tinacms/dist/edit-state'
 We are going to update our props from `({ post, morePosts, preview })` to `(props)`. We are going to pass the props into our `useTina` hook. Now that has been updated with the props coming in we can use `useTina` . 
 
 ```diff
+
 - export default function Post({ post, morePosts, preview }) {
 + export default function Post( props ) {
  


content/blog/from-cms-to-contextual.md
```diff
@@ -0,0 +1,424 @@
...
+    fallback: false,
+  }
+}
+```
+
+**Quick break down of `getStaticPaths`**
+
+The `getStaticPaths` code takes the graphql query we created, because it does not require any `variables` we can send down an empty object. In the return functionality we map through each item in the `postsListData.getPostList` and create a slug for each one.
+
+### Creating getStaticProps query
```
spbyrne, line 185: The heading hierarchy is reversed between this and the next heading (should be H2 > H3)

content/blog/from-cms-to-contextual.md
```diff
@@ -0,0 +1,424 @@
...
+
+As you can see, we are reusing our query from before and passing the variables, and data to `useTina`.  This now means we can power our site using contextual editing. Congratulations your site now has superpowers! 
+
+### Update our elements to use Tina data
+
+Now we have access to our Tina powered data we can go through and update all of our elements to use Tina. You can replace each of the `post.` with `data.getPostsDocument.data.` and then replace the `!post?.slug` with `!props.variables.relativePath` when you are finished your return code should look like:
+
+```jsx
+const router = useRouter()
+  if (!router.isFallback && !props.variables.relativePath) {
```
spbyrne, line 343: This area was a little confusing because you're changing 3 things in close proximity but you don't get a complete picture of what that looks like. Hard to understand where `useTina` goes relative to the other changes you're making. Maybe provide something like this, or a diff view:

```
export default function Post(props) {

  const { data } = useTina({
    query,
    variables: props.variables,
    data: props.data,
  }) 

  const router = useRouter()
  if (!router.isFallback && !props.variables.relativePath) {
    return <ErrorPage statusCode={404} />
  }
  return (
    <Layout preview={false}>
      <Container>
        <Header />
...
```

content/blog/from-cms-to-contextual.md
```diff
@@ -0,0 +1,424 @@
...
+        )}
+      </Container>
+    </Layout>
+    
+  )
+```
+
+### The final piece of Contextual Editing.
+
+One piece of code that the original was doing was taking the markdown and turning it into HTML. We currently aren’t doing that with our body, the reason is, we have to keep track of changes when using Tina. This is where useState and useEffect come in, first create a content variable that is track by a state:
```
spbyrne, line 379: I don't understand this section. We can't transform markdown because we have to track changes?

spbyrne (approved):
Awesome! All the fixes look great. LGTM 👍

PR #1285, opened 2022-04-06
Adds an i18n guide for TinaCMS

Enigmatical:
... but does a bit more than that.  With all of the major changes happening around TinaCMS, we actually retired `/guides` for a bit.  This PR not only introduces a new guide (for i18n using NextJS), but also revives the `guides` section.

I tried to recreate what existed before, replicating the same structure and building the same configuration (meta.json) files.  Can't promise I succeeded in a complete revival, though.

Closes https://github.com/tinacms/product-roadmap/issues/74


content/guides/tinacms/meta.json
```diff
@@ -0,0 +1,4 @@
+{
+  "title": "Tina CMS",
+  "weight": 100
+}
```
Enigmatical, line 4: I added this file, but I'm not entirely sure what it does.  It doesn't seem to be influencing the nav at all (that is based off the `toc_doc`).

data-api/fetchGuides.ts
```diff
@@ -3,7 +3,7 @@ const fg = require('fast-glob')
 var fs = require('fs')
 var path = require('path')
 
-export default async function fetchDocs() {
+export default async function fetchGuides() {
```
Enigmatical, line 6: I just fixed the name for this function to match its purpose (more for code searching because it is exported by `default`).

now.json
```diff
@@ -445,6 +445,10 @@
       "source": "/guides/nextjs(.*)",
       "destination": "/guides/tina-cloud/add-tinacms-to-existing-site/overview/"
     },
+    {
+      "source": "/guides/tinacms(.*)",
+      "destination": "/guides/"
+    },
```
Enigmatical, line 451: This was resurrected.  I assume we also want to _keep_ the ultimate redirect if everything else falls through?
```js
    {
      "source": "/guides/(.*)",
      "destination": "/docs/setup-overview/"
    },
```

pages/_app.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { useEffect } from 'react'
 import App from 'next/app'
 import Head from 'next/head'
+import Script from 'next/script'
```
Enigmatical, line 4: I was getting a repeated error that `<script>` is no longer allowed in `next/head`.  Decided to fix this to quell those warnings by doing what it said (import `Script` and use that instead).

content/toc-doc.json
```diff
@@ -229,6 +229,17 @@
...
   },
+  {
+    "category": "Guides",
+    "slug": "/guides",
+    "items": [
+      {
+        "id": "/guides/tinacms/adding-internationalization-to-tina/overview/",
+        "slug": "/guides/tinacms/adding-internationalization-to-tina/overview/",
+        "title": "Adding Internationalization to Tina"
+      }
```
jamespohalloran, line 240: Right now, non of the sub-pages are accessible in the sidebar.
I think you'll want to set the `expandOnSelect` to true, and add the "items" list with all the sub-pages.
See: https://github.com/tinacms/tinacms.org/pull/1226/files#diff-6e6909b12b1fddd800e449f17dc6fec70644386d5a6ffdacba915a017195b298L399

now.json
```diff
@@ -445,6 +445,10 @@
       "source": "/guides/nextjs(.*)",
       "destination": "/guides/tina-cloud/add-tinacms-to-existing-site/overview/"
     },
+    {
+      "source": "/guides/tinacms(.*)",
```
jamespohalloran, line 449: We had this before because we had no remaining guides in tinacms (There was a single guide in "guides/tina-cloud"). 
I think you'll want 
```suggestion
      "source": "/guides/tina-cloud(.*)",
```

jamespohalloran:
I left some comments around a few layout specific/site things! The content to me looks great, but I'm almost wondering it this may be easier to read if it's just a single page though (In the overview)?
Navigating to a page like this doesn't really tell you much out of context: https://tinacms-site-next-git-feature-add-i18n-guide-tinacms.vercel.app/guides/tinacms/adding-internationalization-to-tina/update-next-config/

(If you decide to pull it all into one document, then this comment wouldn't apply https://github.com/tinacms/tinacms.org/pull/1285#discussion_r844337945)

content/guides/tinacms/adding-internationalization-to-tina/modify-get-static-paths.md
```diff
@@ -0,0 +1,46 @@
...
+  const paths = [];
+
+  tinaProps.data.getPostsList.edges.map((post) => {
+    locales.map((locale) => {
+      paths.push({
+        params: {
+          filename: post.node.sys.filename,
+          locale,
+        },
+      });
```
jeffsee55, line 38: Shouldn't this be 
```suggestion
      paths.push({
        params: {
          filename: post.node.sys.filename,
        },
        locale
      });
```
https://nextjs.org/docs/advanced-features/i18n-routing#dynamic-routes-and-getstaticprops-pages

jeffsee55 (commented):
This is great, I think you've highlighted the important parts perfectly!

content/guides/tinacms/adding-internationalization-to-tina/modify-get-static-props.md
```diff
@@ -0,0 +1,34 @@
...
+
+Next, we'll want to update `getStaticProps` to include `locale` as part of the `relativePath`.
+
+```js
+/**
+ * /pages/post/[filename].tsx
+ */
+
+// `locale` is provided alongside `params`
+const getStaticProps = async({ params, locale }) {
```
logan-anderson, line 14: Should we add a try / catch here as [we do in the data fetching section](https://tinacms-site-next-git-feature-add-i18n-guide-tinacms.vercel.app/docs/features/data-fetching/#querying-tina-content-in-nextjs) ?

logan-anderson (commented):
Looks really good! Just left a couple of comments about the content. Also, I was wondering if we could also add a section about updating or adding a a route mapper as well for this?

Enigmatical:
> I left some comments around a few layout specific/site things! The content to me looks great, but I'm almost wondering it this may be easier to read if it's just a single page though (In the overview)? Navigating to a page like this doesn't really tell you much out of context: https://tinacms-site-next-git-feature-add-i18n-guide-tinacms.vercel.app/guides/tinacms/adding-internationalization-to-tina/update-next-config/
> 
> (If you decide to pull it all into one document, then this comment wouldn't apply [#1285 (comment)](https://github.com/tinacms/tinacms.org/pull/1285#discussion_r844337945))

I did decide to consolidate the guide into a single markdown file (`guide.md`) which simplified the navigation.

I actually really like how this feels and I wonder if this is how we can start to make our Guides section friendly with TinaCMS (managinging the navigation between steps would make it difficult to adapt to TinaCMS, but a single file is much simpler).

content/guides/tinacms/adding-internationalization-to-tina/modify-get-static-props.md
```diff
@@ -0,0 +1,34 @@
...
+  // compose `filename` with `locale`
+  const relativePath = `${params.filename}.${locale}.mdx`
+
+  const tinaProps = await staticRequest({
+    query: `query($relativePath: String) {
```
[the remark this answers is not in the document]
Enigmatical: I ended up pulling in the flow we use in the starters today:
```js
  const client = ExperimentalGetTinaClient();
  const postsListData = await client.getPostsList();
```
The query, etc. isn't really the focus of the guide - it's mostly just using `locale` and manipulating `paths` and `relativePath`, so this feels cleaner.

Enigmatical:
> Looks really good! Just left a couple of comments about the content. Also, I was wondering if we could also add a section about updating or adding a a route mapper as well for this?

As it turns out, this new iteration (using sub-folders) actually negates the need to make changes to the `RouteMapper` (at least in the example).

now.json
jamespohalloran: Just removed that global redirect It was actually redirect any guide link, (including the new one), back to the docs root

jamespohalloran (approved):

PR #1322, opened 2022-05-05
new client docs

logan-anderson:
Depends on https://github.com/tinacms/tinacms/issues/2628 to be merged first

### DIFF
--- a/content/docs/graphql/read-only-tokens.md
+++ b/content/docs/graphql/read-only-tokens.md
@@ -5,6 +5,10 @@ last_edited: '2022-02-07T18:00:00.000Z'
 
 {{ WarningCallout text="This is an experimental feature and may be slow as we work on performance improvements" }}
 
+<br/>
+
+{{ WarningCallout text="It is highly recommended to  [enable the data layer](/docs/tina-cloud/data-layer/#enabling-the-data-layer) to use this feature. Have any thoughts? Let us know in the chat, or through the [GitHub discussion](https://github.com/tinacms/tinacms/discussions/2811)!" }}
+
 Read-only tokens allow data fetching at runtime without the need for the local GraphQL server. Some use cases include the following:
 
 - Runtime server-side logic in `getServerSideProps`, `getStaticProps` (when fallback is not `false`), etc.
@@ -54,7 +58,6 @@ const apiURL =
     : `https://content.tinajs.io/content/<Your Client ID>/github/${branch}`;
 
 // Token generated on app.tina.io
-export const client = createClient({ url: apiURL, token: "***" });
 export const client = createClient({
     // default values
     url: apiURL,
@@ -194,26 +197,28 @@ const variables = {
   relativePath: 'HelloWorld.md',
 }
 
-function BlogPostPage() {
-  const [initalData, setData] = useState(null)
-  const [isLoading, setLoading] = useState(false)
+export default function BlogPostPage() {
+  const [initialData, setData] = useState(null);
+  const [isLoading, setLoading] = useState(false);
 
   useEffect(() => {
-    const fetchContent = async () =>{
-      const data = await client.request({query, variables})
-      setData(data)
-    }
-    fetchContent()
-  }, [query, JSON.stringify(variables)])
+    const fetchContent = async () => {
+      setLoading(true);
+      const data = await client.request({ query, variables });
+      setData(data?.data);
+      setLoading(false);
+    };
+    fetchContent();
+  }, [query, JSON.stringify(variables)]);
 
-  const { data } = useTina({ query, variables, data: initalData })
+  const { data } = useTina({ query, variables, data: initialData });
 
-  if (isLoading) return <p>Loading...</p>
-  if (!data) return <p>No data</p>
+  if (isLoading) return <p>Loading...</p>;
+  if (!data) return <p>No data</p>;
 
-  return <div>{JSON.stringify(data)}</div>
+  return <div>{JSON.stringify(data)}</div>;
 }
-export default BlogPostPage
+
 ```
 
 ### Next.js `fallback: "blocking"`
@@ -245,29 +250,23 @@ export const getStaticProps = async ctx => {
   const variables = {
     relativePath: ctx.params.slug + '.md',
   }
-  let data = {}
-  let error = false
+  let data;
   try {
-    // use the local client at build time
-    data = await client.request({
-      url: 'http://localhost:4001/graphql',
+    // This will use the URL that was passed to the client
+    const res = await client.request({
       query,
       variables,
     });
+    console.log({ res });
+    data = res?.data;
   } catch (error) {
     // swallow errors related to document creation
   }
+
   if (!data) {
-    error = true;
-  }
-  if (error) {
-    // use read only tokens to get live data
-    data = client.request({query, variables})
-    if (!data) {
-      return {
-        notFound: true,
-      }
-    }
+    return {
+      notFound: true,
+    };
   }
 
   return {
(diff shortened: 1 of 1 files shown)


content/docs/graphql/read-only-tokens.md
```diff
@@ -39,9 +43,94 @@ Wild card matching is supported in the branch names using '\*' to match anything
...
+  query: "Your Graphql query",
+  variables: {...},
+  url: "http//localhost:4001/graphql",
+});
+
+```
+
+```ts
+import { client } from "../pathToTina/.tina/client";
+// over-ride token
```
jamespohalloran, line 120:
```suggestion
// override token
```

content/docs/graphql/read-only-tokens.md
```diff
@@ -39,9 +43,94 @@ Wild card matching is supported in the branch names using '\*' to match anything
...
+    url: apiURL,
+    token: "Your Read Only Token generated above",
+})
+```
+When using "http://localhost:4001/graphql" the content will be queried from file system (This only works during devolvement or in CI) and when using `https://content.tinajs.io/content/<Your Client ID>/github/${branch}` the content will be queried from Tina cloud. 
+
+In most cases the `apiURL` is the same one that is used for editing. So instead of passing the passing `apiURL` the client can be passed.
 
-Now you can make a POST request to the content API with the desired GraphQL request.
+`.tina/schema.{ts,tsx,js}`
```
jamespohalloran, line 71: Might be more readable if this is a comment within the below code snippet

content/docs/graphql/read-only-tokens.md
```diff
@@ -39,9 +43,94 @@ Wild card matching is supported in the branch names using '\*' to match anything
...
+import { client } from "../pathToTina/.tina/client";
+
+const data = await client.request({
+  query: "Your Graphql query",
+  variables: {...},
+});
+
+```
+
+*optionally* the `url` or `token` can be overridden from request to request.
```
jamespohalloran, line 105: Is the only case where we can see this being overridden only for static with the filesystem? I almost wonder if this belongs more in a reference section.

jamespohalloran (commented):
I think this doc is good for now! But we may want to think about breaking it up in the future. The `curl` & `fetch` content feels like it may belong in a reference doc for our content api. 
The next-specific fallback blocking stuff maybe feels like it could be its own guide specific to ISR?

content/docs/graphql/read-only-tokens.md
logan-anderson: I reference section would be good for this. Only reason I did not include it was because I did not want users to see that before reading this doc.

content/docs/graphql/read-only-tokens.md
```diff
@@ -39,9 +43,94 @@ Wild card matching is supported in the branch names using '\*' to match anything
...
+// Token generated on app.tina.io
+export const client = createClient({
+    // default values
+    url: apiURL,
+    token: "Your Read Only Token generated above",
+})
+```
+When using "http://localhost:4001/graphql" the content will be queried from file system (This only works during devolvement or in CI) and when using `https://content.tinajs.io/content/<Your Client ID>/github/${branch}` the content will be queried from Tina cloud. 
+
+In most cases the `apiURL` is the same one that is used for editing. So instead of passing the passing `apiURL` the client can be passed.
```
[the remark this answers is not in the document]
logan-anderson: I think we should continue documenting passing `apiURL`  until we make using the data layer the default. 

Right now, using and passing the client is awkward if the user does not have the data layer enabled. (The will hit Github's Api limits)

logan-anderson:
> I think this doc is good for now! But we may want to think about breaking it up in the future. 

> The `curl` & `fetch` content feels like it may belong in a reference doc for our content api. 

I can move this.

> The next-specific fallback blocking stuff maybe feels like it could be its own guide specific to ISR?

I think this is a great idea and could be addressed in a separate ticket?

content/docs/graphql/read-only-tokens.md
```diff
@@ -39,9 +43,94 @@ Wild card matching is supported in the branch names using '\*' to match anything
 
 Wild card matching is useful for matching branches that have not been created yet and can be used for editorial workflows.
 
-### Make a fetch request using the API key
+### Making requests with the tina client
+#### Setting up the Tina client
+
+This client can be used for data fetching with Tina. It can be used on the client and server to make graphQL requests. 
+
+To set up first add a new file called `.tina/client.{js,ts}`
```
Enigmatical, line 51:
```suggestion
To set up, first add a new file called `.tina/client.{js,ts}`
```

content/docs/graphql/read-only-tokens.md
```diff
@@ -39,9 +43,94 @@ Wild card matching is supported in the branch names using '\*' to match anything
...
+    : `https://content.tinajs.io/content/<Your Client ID>/github/${branch}`;
+
+// Token generated on app.tina.io
+export const client = createClient({
+    // default values
+    url: apiURL,
+    token: "Your Read Only Token generated above",
+})
+```
+When using "http://localhost:4001/graphql" the content will be queried from file system (This only works during devolvement or in CI) and when using `https://content.tinajs.io/content/<Your Client ID>/github/${branch}` the content will be queried from Tina cloud.
```
Enigmatical, line 67:
```suggestion
When using "http://localhost:4001/graphql," the content will be queried from file system (This only works during devolvement or in CI) and when using `https://content.tinajs.io/content/<Your Client ID>/github/${branch}` the content will be queried from Tina cloud. 
```

content/docs/graphql/read-only-tokens.md
Enigmatical, line 69:
```suggestion
In most cases, the `apiURL` is the same one that is used for editing. So instead of passing the passing `apiURL` the client can be passed.
```

content/docs/graphql/read-only-tokens.md
```diff
@@ -39,9 +43,94 @@ Wild card matching is supported in the branch names using '\*' to match anything
...
+export const tinaConfig = defineConfig({
++  client,
+-  url: ...
+  //...
+})
+```
+
+#### Requesting data using Read only tokens
+
+Now that the client is defined we can use it to query our content.
```
Enigmatical, line 85:
```suggestion
Now that the client is defined, we can use it to query our content.
```

content/docs/graphql/read-only-tokens.md
```diff
@@ -39,9 +43,94 @@ Wild card matching is supported in the branch names using '\*' to match anything
...
+});
+
+```
+
+### Making requests with `curl` and `fetch`
+
+> NOTE: for most cases the tina client can be used and it is not necessary to use fetch directly
 
-Now you can make a POST request to the content API with the desired GraphQL request.
+If you do not want to use the Tina client you can make a POST request directly to the content API.
```
Enigmatical, line 133:
```suggestion
If you do not want to use the Tina client, you can make a POST request directly to the content API.
```

Enigmatical (approved):
I think I agree with James that there's quite a lot here, but I'm not sure exactly the best approach to break it up (maybe once the `client` is the "all client" and we can write a whole section on its many different uses.

Otherwise, this looks good.  I added a few commas where they were needed 👍

content/docs/graphql/read-only-tokens.md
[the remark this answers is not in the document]
jamespohalloran: > I think we should continue documenting passing apiURL until we make using the data layer the default.

> Right now, using and passing the client is awkward if the user does not have the data layer enabled. (The will hit Github's Api limits)

We could just override the URL in getStaticProps / paths though right? 

If we want the experimental client contained as experimental, I just think it's fine keeping it contained here. It just caused me to have to think a little bit harder since the read-only-tokens, and the client are 2 new concepts outlined in this doc.

jamespohalloran:
> > The next-specific fallback blocking stuff maybe feels like it could be its own guide specific to ISR?
> 
> I think this is a great idea and could be addressed in a separate ticket?

Yep that's fine by me!

jamespohalloran (approved):

PR #1326, opened 2022-05-10
move playgrounds to external links

jamespohalloran:
Our playgrounds are pretty heavy browser loads, so pages that feature multiple playgrounds should instead link to external pages

### DIFF
--- a/content/docs/schema.md
+++ b/content/docs/schema.md
@@ -60,7 +60,7 @@ This is my main post body.
 
 Once we've defined a collection, we can edit its fields through the Tina UI, or [query its content](/docs/graphql/overview/) using the Tina Content API.
 
-<iframe width="100%" height="450px" src="https://tina-gql-playground.vercel.app/string-body" />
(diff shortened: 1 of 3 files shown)


content/docs/schema.md
```diff
@@ -60,7 +60,7 @@ This is my main post body.
 
 Once we've defined a collection, we can edit its fields through the Tina UI, or [query its content](/docs/graphql/overview/) using the Tina Content API.
 
-<iframe width="100%" height="450px" src="https://tina-gql-playground.vercel.app/iframe/string-body" />
+<iframe width="100%" height="450px" src="https://tina-gql-playground.vercel.app/string-body" />
```
Enigmatical, line 63: Just to confirm, did you mean for this one to remain an `iframe` or switch to an `a`?

jamespohalloran: oops good catch, just switched that back to use the iframe url

jamespohalloran: But yeah, i was thinking having one iframe per page is fine, and gives a "taste" of the examples

Enigmatical: I like that.  👍

PR #1342, opened 2022-05-24
Update cta

perkinsjr:
Updates the CTA Button to use both a configurable URL and Button text.

### DIFF
--- a/pages/blog/[slug].tsx
+++ b/pages/blog/[slug].tsx
@@ -37,10 +37,10 @@ const components = {
       allowFullScreen={true}
     ></iframe>
   ),
-  CreateAppCta: ({ ctaText, cliText }) => (
+  CreateAppCta: ({ ctaText,ctaURL, cliText }) => (
     <>
       <a
-        href={ctaText}
+        href={ctaURL}
         style={{
           textDecoration: 'none',
           borderRadius: '10px',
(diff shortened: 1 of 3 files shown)


pages/blog/[slug].tsx
```diff
@@ -37,10 +37,10 @@ const components = {
       allowFullScreen={true}
     ></iframe>
   ),
-  CreateAppCta: ({ ctaText, cliText }) => (
+  CreateAppCta: ({ ctaText,ctaURL, cliText }) => (
```
Comment (jamespohalloran) [inline pages/blog/[slug].tsx:40]: If the URL is configurable, should this be renamed to something more generic?
(E.g CTAButton)

jamespohalloran (changes requested):

pages/blog/[slug].tsx
Comment (perkinsjr) [reply pages/blog/[slug].tsx:40]: I add it to be config so that we don't have to manually update when we update the docs URL in X months time

Comment (jamespohalloran) [reply pages/blog/[slug].tsx:40]: > I add it to be config so that we don't have to manually update when we update the docs URL in X months time

I'm not sure what you mean here. If we change the docs URL, wouldn't that mean needing to change every instance of CreateAppCta instead of changing the one link within `CreateAppCta`? 

Not a big deal though, but just felt odd that the name `CreateAppCta` feels very specific to "Creating an app", however, the link is configurable.

Comment (jamespohalloran) [reply pages/blog/[slug].tsx:40]: If you want to keep it configurable though, you'll also have to set it on the other places we use `CreateAppCta`

jamespohalloran (approved):

PR #1364, opened 2022-06-29
New Media docs

jamespohalloran:
- [x] DO NOT MERGE UNTIL REPO-BASED MEDIA IS TESTED & RELEASED

Add repo-based media docs. Update Cloudinary media docs to use config in the defineSchema object


logan-anderson:
Just reading thew this and wondering if we should mention some of the current behaviours / "caveats ".
- When you add new media it gets saved to the default branch (I believe this is the behaviour @kldavis4  ? ) 
- If you are using Tina-cloud you will get the assets from us and not a relative path

There may be a couple others to mention as well.

kldavis4:
> Just reading thew this and wondering if we should mention some of the current behaviours / "caveats ".
> 
> * When you add new media it gets saved to the default branch (I believe this is the behaviour @kldavis4  ? )
> * If you are using Tina-cloud you will get the assets from us and not a relative path
> 
> There may be a couple others to mention as well.

The branch sync logic is currently:
- if you only have a single branch with media enabled, that is the branch that will get the media (when added via the media store)
- if you have multiple branches with media enabled, then the default branch will get that media

PR #1376, opened 2022-08-02
update schema docs

logan-anderson:
cc @jamespohalloran

### DIFF
--- a/content/docs/tina-cloud/connecting-site.md
+++ b/content/docs/tina-cloud/connecting-site.md
@@ -68,8 +68,8 @@ import { defineSchema } from 'tinacms'
 const branch =
   process.env.NEXT_PUBLIC_TINA_BRANCH ||
   process.env.NEXT_PUBLIC_VERCEL_GIT_COMMIT_REF ||
-  process.env.HEAD ||
-  ''
+  process.env.HEAD
+
 const schema = defineSchema({
   config: {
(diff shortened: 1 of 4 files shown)


content/docs/tina-cloud/connecting-site.md
```diff
@@ -68,8 +68,8 @@ import { defineSchema } from 'tinacms'
 const branch =
   process.env.NEXT_PUBLIC_TINA_BRANCH ||
   process.env.NEXT_PUBLIC_VERCEL_GIT_COMMIT_REF ||
-  process.env.HEAD ||
-  ''
```
jamespohalloran, line 72: So the reason I had this set as an empty string as a fallback, is because `branch` is currently required, however, the environment variables may be undefined. (I believe our client generation still catches this and throws an error on empty string).

Otherwise, if you're using typescript, you need to do: something like: 
```
   branch:
      process.env.NEXT_PUBLIC_TINA_BRANCH! || 
      process.env.NEXT_PUBLIC_VERCEL_GIT_COMMIT_REF! || 
      process.env.HEAD!
```
which doesn't work on `.js`.

Open to ideas on a best practice for setting this branch variable though!

logan-anderson: Makes sense! I will add the `|| ''` back in

jamespohalloran (approved):

PR #1384, opened 2022-08-15
Add notes on handling markwon

jeffsee55:
### General Contributing:

* [x] Have you followed the guidelines in our [Contributing document](https://github.com/tinacms/tinacms.org/blob/master/CONTRIBUTING.md)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?


content/docs/editing/mdx.md
```diff
@@ -125,6 +125,160 @@ type TinaMarkdown = ({
...
+            },
+          ],
+        },
+      ],
+    },
+  ]
+}
+```
+
+### Workarounds
```
jamespohalloran, line 239: It's hard to know from the header, or without reading the full document what this is a workaround for. 
Ideally we want to make these headings as skim'able as possible.

jeffsee55:
> I do also kind of wonder if some of the content feels a bit more reference-level

Ok let me think about this and move it somewhere, I mainly just want to get this information out there so we can point to it.

PR #1400, opened 2022-09-16
Rework docs to datafetch with tina

jamespohalloran:
Starting to update the Tina implementation to use the latest paterns rather than fetching with fs. Lots more to update

TODO: 
- [ ] Fix our TOS sidebar nav in docs
- [ ] Use get<Collection>Connection for all page-types
- [ ] Make sure SEO is scraped from rich-text correctly
- [ ] Move layout-level stuff to wrapper
- [ ] Fix any files with unsupported elements (e.g shortcodes)


components/toc/index.tsx
```diff
@@ -36,13 +37,44 @@ const Toc = ({ tocItems, activeIds }: TocProps) => {
...
       </TocContent>
     </TocWrapper>
   )
 }
 export default Toc
 
+const TOCTitle1 = props => {
+  return (
+    <li>
+      {/* <a href={`#${props.children.props.content[0].text.replace(' ', '-')}`}> */}
```
bikkimahato, line 59: Commented code should not be pushed to GitHub.

components/toc/index.tsx
```diff
@@ -36,13 +37,44 @@ const Toc = ({ tocItems, activeIds }: TocProps) => {
...
+      {/* </a> */}
+    </li>
+  )
+}
+
+const TOCTitle2 = props => {
+  return (
+    <ul>
+      <li>
+        {/* <a href={`#${props.children.props.content[0].text.replace(' ', '-')}`}> */}
```
bikkimahato, line 70: Same as above.

PR #1401, opened 2022-09-19
Update markdown docs

jeffsee55:
- Removes outdate markdown doc
- Updates to "concept" doc for markdown/mdx to be less reference-like
- Updates `rich-text` doc to include information related to handling markdown and some more advanced usage patterns.

closes https://linear.app/tina/issue/ENG-478/update-markdown-docs


content/docs/editing/mdx.md
```diff
@@ -1,167 +1,51 @@
 ---
-title: Rich text with MDX
+title: Markdown & MDX
```
jamespohalloran, line 2: Might make more sense to have the name of this file be "markdown" rather than "mdx"?

jamespohalloran (changes requested):
I made some suggestions here:
https://github.com/tinacms/tina.io/pull/1403

We ideally want to give the user enough to get started without having to dig too deep into the reference docs, so I made some updates to make the markdown/mdx doc a bit mroe self-contained

PR #1402, opened 2022-09-20
Logan/eng 507 reference documentation for graphql

logan-anderson:
Adds docs for the shape of a "Node"

Fixes https://linear.app/tina/issue/ENG-507/reference-documentation-for-graphql-nodes


jamespohalloran (approved):
Looks good!

PR #1403, opened 2022-09-21
Update markdown revisions

jamespohalloran:
I found the existing markdown/mdx docs maybe assumed the user has a bit more context than they have/need.
This turns that doc into a bit more a self-contained "guide", and is a bit more focused on getting the user started.


content/docs/editing/mdx.md
```diff
@@ -4,48 +4,228 @@ id: '/docs/editing/mdx'
...
+## Knock Knock!
+
+Who's there?
+```
+
+The fields registered in between the `---` delimiters, are called your **frontmatter**. Anything below is your **markdown body**.
+
+### What is MDX?
+
+MDX is similar to Markdown, however is adds the ability to write JSX components in your **markdown body**.
```
jeffsee55, line 31:
```suggestion
MDX is similar to Markdown, however it adds the ability to write JSX components in your **markdown body**.
```

PR #1413, opened 2022-10-03
added defaultItem docs

logan-anderson:
fixes ENG-598
fixes https://linear.app/tina/issue/ENG-598/list-true-not-fully-documented-in-reference-docs

### DIFF
--- a/content/blog/tina-version-0.69.13.md
+++ b/content/blog/tina-version-0.69.13.md
@@ -82,7 +82,7 @@ const schema = defineSchema({
       ui: {
         filename: {
           // if disabled, the editor can not edit the filename
-          disabled: true,
+          readonly: true,
           // Example of using a custom slugify function
           slugify: values => {
             // Values is an object containing all the values of the form. In this case it is {title?: string, topic?: string}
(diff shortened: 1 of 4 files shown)


jamespohalloran:
Does `defaultItem` & `list: true` need to be added to other field types? I assumed that you could have a list of numbers for instance, but maybe that's not supported?

logan-anderson:
> Does `defaultItem` & `list: true` need to be added to other field types? I assumed that you could have a list of numbers for instance, but maybe that's not supported?

I think it is supported in the backend but not by our UI

jeffsee55:
>  I assumed that you could have a list of numbers for instance, but maybe that's not supported?

You can but the list of items you can do with that is incomplete I believe, I think `string`, `number`, maybe `boolean`, but then not `rich-text` or `image`

jamespohalloran:
> > I assumed that you could have a list of numbers for instance, but maybe that's not supported?
> 
> You can but the list of items you can do with that is incomplete I believe, I think `string`, `number`, maybe `boolean`, but then not `rich-text` or `image`

I just checked, it does look like it's only supported through the UI on strings & objects, so the docs are correct.

jamespohalloran (approved):

PR #1416, opened 2022-10-07
Feat/iframe docs

logan-anderson:
WIP of updating the docs to make the `iframe` mode the "happy path"


logan-anderson:
The upgrade blog post looks good @jamespohalloran It might be good to add a node that everything in `schema.config` should be moved to defineConfig({...})

PR #1434, opened 2022-11-21
Bug fixes logan

logan-anderson:
cc @spbyrne


logan-anderson:
I think everything is working except for short-codes (which are not supported anymore in react-markdown)

PR #1440, opened 2022-11-23
Recent Posts Block

spbyrne:
Work-in-progress page block for displaying recent posts.

Screen Shot 2022-11-23 at 5 07 16 PM


components/blocks/RecentPosts.tsx
```diff
@@ -23,30 +22,31 @@ export const RecentPostsBlock = ({ data, index, recentPosts }) => {
...
-                  </MetaBit>
-                </BlogMeta>
-                {getExcerpt(post.body, 200)}
-              </RichTextWrapper>
-            </a>
-          </DynamicLink>)
-          })
-          }
+          {recentPosts.edges.map(({ node: post }) => {
+            const slug = post['_sys']['filename']
```
jamespohalloran, line 26:
```suggestion
const slug = post._sys.filename
```
^ Can you do this? It will give you better typing

spbyrne: Sure, looks a little cleaner. I'm not getting any type hints either way, though.

PR #1448, opened 2022-12-05
add frameworks documentation

Phoenix-Alpha:
[ENG-689](https://linear.app/tina/issue/ENG-689/provide-more-guidance-for-public-assets-config-prompt)


scottgallant:
@Phoenix-Alpha can we use a table instead of headings and paragraphs?

jamespohalloran:
I think this doc would be better suited lower in the docs sidebar (maybe underneath a "Frameworks" section. The "Getting Started" section of the docs is treated more as a tutorial / intro to Tina, and users may not have a lot of context from this doc alone. 

Screen Shot 2022-12-06 at 8 54 37 AM

I think Netlify does a good job with their docs for this similar type of thing: https://docs.netlify.com/integrations/frameworks/

They have a high level doc that details the frameworks supported, their build commands, etc, but they also have individual sub-docs for each section (which relates to: https://linear.app/tina/issue/MAR-8/framework-specific-guides)

jamespohalloran (approved):

PR #1450, opened 2022-12-06
Docs Nav Improvements

spbyrne:
This imprroves the docs nav UX by adding accordions, better active link highlighting, and scrolling to a lower link on load so that it's in view. It also simplifies the data. 

By default only the first two levels of nav items are displayed, but even top level categories can be minimized by the user.

Screenshot 2022-12-06 at 3 55 31 PM

Any item with sub items acts as an accordion toggle (not a link), so any items that didn't already have a child with the same slug have an 'Overview' item added.

Screenshot 2022-12-06 at 3 56 57 PM

The nav will also scroll on load so that the current nav link is visible:

Screenshot 2022-12-06 at 3 58 23 PM

Unrelated to the docs nav I fixed some React warnings.


spbyrne:
Actually, I think parents of nested selected elements need some sort of highlighting, I'll add that.

PR #1470, opened 2023-01-13
added docs for forestty migration updates

logan-anderson:
This is an update to the docs that guides the users though using the forestry migration tool, the common errors that may occur and how to fix them. This also updates the migration doc to assume that the user will use this tool.


content/docs/forestry/common-errors.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Error when getting content in the admin
+
+```
+[Error] GetCollection failed: Unable to fetch, errors: Error querying file <File> from collection pages. Please run "tinacms audit" or add the --verbose option for more info
+```
+
+![](https://res.cloudinary.com/forestry-demo/image/upload/v1673619483/tina-io/docs/forestry-migration/Screen-Shot-Error-Messager.png)
+
+This can happen for many reason but a common reason is because the data you are trying to fetch is not in the correct format, missing a required field or is missing the `_template` key. If you look at the error message in the terminal you should see the exact error that is causing the issue.
```
jamespohalloran, line 95:
```suggestion
This can happen for many reasons. Some common reasons are because the data you are trying to fetch is 
- not in the correct format 
- missing a required field
- is missing the `_template` key. 
If you look at the error message in the terminal you should see the exact error that is causing the issue.
```

content/docs/forestry/common-errors.md
```diff
@@ -0,0 +1,134 @@
...
+This can be fixed by adding the `_template` key to the front matter of the file.
+
+```md
+---
+# ...
+_template: my_template
+---
+```
+
+Where "my_template" is the name of the template in the collection to which the file belongs.
```
jamespohalloran, line 113:
```suggestion
..where "my_template" is the name of the template in the collection to which the file belongs.
```

jamespohalloran (changes requested):
This is super helpful! Nice job @logan-anderson

jamespohalloran (approved):

PR #1471, opened 2023-01-16
fix issue with mdx

logan-anderson:
There was an issue with this blogs markdown file that causes a render error.


logan-anderson:
This is what is looks like right now


Screen Shot 2023-01-16 at 10 13 13 AM

PR #1522, opened 2023-03-28
Remove common errors section relating to noSDK option

jeffsee55:
I think we can skip this once the skip client config lands. Depends on https://github.com/tinacms/tinacms/pull/3734


logan-anderson:
I think this _might_ be valuable somewhere in the docs but I agree it can be removed from the Forestry migration section.

PR #1559, opened 2023-05-16
Quick edit docs

jeffsee55:
https://tina-io-git-quick-edit-tinacms.vercel.app/docs/editing/quick-editing/

FYI This leaves out a couple of things:

A bit about the `data-tina-field-overlay` support, the starter doesn't have a good need for it so it's hard to demonstrate

The Vercel preview docs for now, that probably is best to just live as a README in the monorepo.

### DIFF
--- a/content/docs/editing/quick-editing.md
+++ b/content/docs/editing/quick-editing.md
@@ -15,6 +15,20 @@ last_edited: '2021-11-06T18:00:00.000Z'
 Tina's quick-editing feature allows editors to select the element they want
 to edit on the page in order to see it in the sidebar.
 
+<video
+className="video"
+autoPlay="true"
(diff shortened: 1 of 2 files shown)


content/docs/editing/quick-editing.md
```diff
@@ -0,0 +1,166 @@
...
+            </a>
+          </li>
+        ))}
+      </ul>
+    </div>
+  )
+}
+```
+
+Notice that the link data attribute only needs access to the `link` data object.
```
jeffsee55, line 114:
```suggestion
Notice that the `<a>` tag's data attribute only needs access to the `link` data object.
```

jamespohalloran (approved):

PR #1569, opened 2023-05-31
Rm old active field indicator docs

jamespohalloran:
Remove old active field indicator docs


jamespohalloran:
> @jamespohalloran does the menu need to be updated to remove the link to the active field indicator doc?

whoops! Yes it does. updated

PR #1601, opened 2023-07-04
ai-markdown blog tweaks

jamespohalloran:
Some of these are stylistically a bit different, so wanted to open up a PR into yours incase we want to throw out these suggestions. 
A few main tweaks I've added:
- A little bit of an intro into why TinaCMS is a great fit for leveraging AI (in terms of its custom workflows)
- Mention some of the limitations of AI, as it can be a bit touchy of a subject.


logan-anderson:
Looks good to me! Feel free to merge this in @jamespohalloran

PR #1649, opened 2023-09-05
docs: Improve docs for non-next external-media

jamespohalloran:
Improve docs for non-next external-media implementations.

One thing to confirm, I removed the NEXT_PUBLIC_ prefix from our environment variables. It doesn't look like these variables are refenced in the tinacms packages, and I don't believe they are needed on the frontend, but could use a second look just to confirm I'm not missing something.


kldavis4 (approved):
lgtm

PR #1693, opened 2023-11-15
auth provider reference docs refactor

kldavis4:
- changed authentication -> authprovider wherever needed
- moved backend hosting to reference docs
- made the reference docs less neutral so its clearer that the recommended auth provider is authjs


logan-anderson (approved):

logan-anderson:
I left one comment on vercel but looks good overall.

PR #1695, opened 2023-11-16
add reference doc for switch to discord auth provider

kldavis4:


logan-anderson:
@kldavis4 Just want to confirm something.

Even with this flow I still need to add users in the CMS correct? The users discord username must be in the users collection correct?

kldavis4:
> @kldavis4 Just want to confirm something.
> 
> Even with this flow I still need to add users in the CMS correct? The users discord username must be in the users collection correct?

yeah, that's covered in this update

PR #1774, opened 2024-05-20
change netlifytoDecap

joshbermanssw:
netlify cms is now called decap cms


leahy268 (approved):
Just testing if this fixes the main/master branch problem

Issue #1780, opened 2024-05-20
💄 Tina.io | CTA Button Styling 


joshbermanssw:
As per our call @sethdaily and the email : Re: Tina.io - add a call to action next to the npx command

- [ ] Add two CTA buttons linked to the appropriate destination 
- [ ] Make sure buttons are on a newline 
- [ ] Make sure there is an 'or' between the 'Book a demo' button and the npx command

joshbermanssw:
Done 
![Screenshot 2024-05-22 at 10 16 35 am](https://github.com/tinacms/tina.io/assets/137844305/9a6d9c53-b04e-4549-b6e3-e2a0007e8e8a)
**Figure: Live site with updated UI **

PR #1781, opened 2024-05-22
CTA Styling 

joshbermanssw:
As per @adamcogan , @sethdaily and @bradystroud #1780  was created to fix the styling of the CTA buttons added to the homepage

**Changes**
- Buttons now sit on new lines 
- 'or' text between the buttons and the npx command


bradystroud (approved):
LGTM

PR #1787, opened 2024-05-22
Adding Ids on Action Blocks for Google Tag Manager

joshbermanssw:
As per #1786 i am assigning ids for each action so that we can track them in Google Tag Manager.

 We have also modified the logic to do with action lists so that the list styling applied to the CTA buttons do not ruin the formatting of the rest of the site. This is so that groups of action buttons > 2 will be stacked vertically rather than horizontally.

### DIFF
--- a/components/blocks/Actions.tsx
+++ b/components/blocks/Actions.tsx
@@ -122,11 +122,12 @@ export const Actions = ({ items, align = 'left' }) => {
 
 export const CodeButton = ({ children, ...props }) => {
   const [copied, setCopied] = React.useState(false)
+  const id = sanitizeLabel(children.toString())
 
   const clickEvent = () => {
     setCopied(true)
     copyToClipboard(children)
-    setTimeout(() => {
+        setTimeout(() => {
       setCopied(false)
     }, 2000)
   }
@@ -137,7 +138,7 @@ export const CodeButton = ({ children, ...props }) => {
         className="button event-cmd-button"
         onClick={clickEvent}
         {...props}
-        id={sanitizeLabel(children.toString())}
+        id={id}
       >
         <span className={`success-message ${copied ? `visible` : ``}`}>
           Copied to clipboard!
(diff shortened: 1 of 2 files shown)


components/blocks/Actions.tsx
```diff
@@ -101,11 +122,12 @@ export const Actions = ({ items, align = 'left' }) => {
 
 export const CodeButton = ({ children, ...props }) => {
   const [copied, setCopied] = React.useState(false)
+  const id = sanitizeLabel(children.toString())
 
   const clickEvent = () => {
     setCopied(true)
     copyToClipboard(children)
-    setTimeout(() => {
+        setTimeout(() => {
```
JackDevAU, line 130: small one but can we fix the indentation 😅

components/blocks/Actions.tsx
```diff
@@ -101,11 +122,12 @@ export const Actions = ({ items, align = 'left' }) => {
 
 export const CodeButton = ({ children, ...props }) => {
   const [copied, setCopied] = React.useState(false)
+  const id = sanitizeLabel(children.toString())
```
JackDevAU, line 125: Instead of using the children which can be anything - instead can we pass down some unique value (if thats whats intended)

components/ui/Button.tsx
```diff
@@ -66,6 +67,7 @@ export const LinkButton = ({
   children,
   ...props
 }) => {
+  const id = sanitizeLabel(children[0].toString());
```
JackDevAU, line 70: Instead of using the children which can be anything - instead can we pass down some unique value (if thats whats intended)

components/blocks/Actions.tsx
joshbermanssw: made it children[0] (the label of the object)

components/ui/Button.tsx
```diff
@@ -1,3 +1,4 @@
+import { sanitizeLabel } from 'components/blocks'
```
JackDevAU, line 1: may as well remove this :)

JackDevAU (approved):

Issue #1798, opened 2024-05-24
✨ Home page - blog post removing link text


landonmaxwell:
### 🩸 Pain
When a blog post has a url link in it, the link isn't showing up in the Recent Posts blog preview on the home page.
Instead it completely removes it, making sentences look incorrect.

### 💼 Tasks 

- [x] Stop the hyperlink from disappearing

### ✅ Acceptance Criteria 

- [x] Checked by William L & Landon

### 📸 Screenshots 

![image](https://github.com/tinacms/tina.io/assets/78135895/663ee9fb-bf9c-448f-b021-82b73c661047)

![image](https://github.com/tinacms/tina.io/assets/78135895/b204e447-f434-4bc6-84de-6bfe3145f230)

Cc: 
@william-liebenberg @sethdaily @wicksipedia

joshbermanssw:
<img width="719" alt="Screenshot 2024-05-27 at 10 58 59 am" src="https://github.com/tinacms/tina.io/assets/137844305/d6ee1adb-7be4-414a-822a-70fc7dc40668">

**Figure: PR mentioned above, includes <a> text in blog posts**

joshbermanssw:
<img width="719" alt="Screenshot 2024-05-27 at 10 58 59 am" src="https://github.com/tinacms/tina.io/assets/137844305/91741910-3ea7-4f9e-bdf5-579ef35d6b68">

**Figure: Fixed - request was only looking for 'text' text, not other types (i.e hyperlink)** 

✅Test pass with Landon

PR #1801, opened 2024-05-27
checking more types of text in payload

joshbermanssw:
As per #1798 the blog posts were not including text that was not of type 'text'. Therefore all text of type <a> was being dismissed. This change modified the `getExcerpt` method used to retrieve text from posts, making sure we include anchored text and map it to text.

Screenshot 2024-05-27 at 10 58 59 am

**Figure: After the change has been implemented, <a> text now appears**

### DIFF
--- a/utils/getExcerpt.ts
+++ b/utils/getExcerpt.ts
@@ -5,19 +5,24 @@ export const getExcerpt = (
   return body.children
     .filter((c) => c.type == 'p')
     .reduce(function (excerpt, child) {
-      // combine all of child's text nodes into a single string
+      // combine all of child's text and link nodes into a single string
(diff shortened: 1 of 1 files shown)


JackDevAU:
Just for clarification - this PR turns <a> tags inside content that weren't previously displayed into normal text and removes the link so it isn't clickable?

joshbermanssw:
> Just for clarification - this PR turns tags inside content that weren't previously displayed into normal text and removes the link so it isn't clickable?

Yes, for the 'preview' screen on our 'Recent Posts' section. Before it was grabbing text that was strictly 'text' type

JackDevAU (approved):

Issue #1802, opened 2024-05-29
✨ Google Tag Manager - Add Tags to Tina Feature Boxes 


joshbermanssw:
### 🩸 Pain

We want to change the design of the tina.io homepage, but we dont know if users currently use the links on these. 

### 💡 Suggested Solution

Add tags so we can track statistics on Google Analytics and compare with future iterations of the homepage and specifically the feature box re-vamp. 

We also want to know which features users are most interested in. 

### 💼 Tasks 

- [x] Add individual unique tags to each of the 8 boxes
- [x] Confirm these can be viewed in Google Analytics 

### ✅ Acceptance Criteria 

- [x] Google Analytics shows user interaction with the 8 feature boxes 

### 📸 Screenshots 

Cc:

joshbermanssw:
![Screenshot 2024-05-31 at 10 06 27 am](https://github.com/tinacms/tina.io/assets/137844305/1bf139a3-e7dc-458e-807b-471d0fd16899)

**Figure: Tags are successfully firing in GTM**

PR #1811, opened 2024-05-30
Feature Grid Id for GTM 

joshbermanssw:
As per #1802 we have added id's to the feature boxes so that we can collect data for user traffic through their redirect links, this will be useful for marketing teams and to compare to future implementations


joshbermanssw:
Testing complete - Tags are fired and picked up in GTM 
Screenshot 2024-05-31 at 10 06 27 am

JackDevAU (approved):

PR #1828, opened 2024-06-05
💄 Updated OG Image

joshbermanssw:
As per email Re: Tina.io open graph OG
cc: @adamcogan @betty

We have upgraded the OG image


pages/_app.tsx
```diff
@@ -45,7 +45,7 @@ const MainLayout = ({ Component, pageProps }) => {
           site_name: data.title,
           images: [
             {
-              url: 'https://tinacms.org/img/tina-twitter-share.png',
+              url: './public/img/tina-og.png',
```
JackDevAU, line 48: does this need the `./public` ?

pages/_app.tsx
```diff
@@ -45,7 +45,7 @@ const MainLayout = ({ Component, pageProps }) => {
           site_name: data.title,
           images: [
             {
-              url: 'https://tinacms.org/img/tina-twitter-share.png',
```
bradystroud, line 48: Is the old image stored on this repo? If so you can remove it :)

joshbermanssw: ✅

pages/_app.tsx
joshbermanssw: i would imagine so

JackDevAU: https://www.opengraph.xyz/url/https%3A%2F%2Ftina-jx24qfqr2-tinacms.vercel.app%2F

Seems like somethings wrong

joshbermanssw: yes - was using path dir not url

JackDevAU (approved):

PR #1827, opened 2024-06-05
💄 Homepage - Logo Carousel 

joshbermanssw:
As per #1809 we wanted to create a logo carousel to display all of the companies using tina!

Design has been test passed by @bettybondoc

Changes
- [ ] Added slider component to logogrid.tsx file 
- [ ] added svgs for respective logos

https://github.com/tinacms/tina.io/assets/137844305/380fbad5-2035-49a3-8697-7a81bb2f0886

PR #1835, opened 2024-06-07
Added IDs : Logo Grid -> Showcase

joshbermanssw:
Apart of #1827 was to link from logos in the carousel to their respective position on the showcase back. This was achieved by creating IDs per showcase section and link that from the carousel item


bradystroud (approved):
LGTM

PR #1847, opened 2024-06-12
🐛 Nav Bar Flicker Bug 

joshbermanssw:
cc: @bradystroud

As per email from @adamcogan

(Checked by Marcus, Eve)

Heya Josh 👋
https://youtu.be/Vagxix6k3ik?si=EdRsEkVCE1vU_vm2

Why at 14:29 does it flicker?
Can you fix?
This PR addresses this issue, re-working how the nav bar fixed mode is set. This was the issue causing the flickering behaviour


bradystroud (approved):
Approving because it fixes the problem, but i dont think its the best solution...

Issue #1840, opened 2024-06-11
🐛 Homepage UI Issues


bettybondoc:
### 🐛 Describe the Bug
UI issues on desktop and mobile view

### 🧑‍💻 Reproduction Steps
1. Go to tina.io on mobile and desktop
2. See UI issues

### 🎯 Expected Behaviour 
Clean UI with enough spacing between elements and consistent alignment/font size

### 💼 Tasks

- [x] Mobile - Add spacing between GIF and heading (see screenshots)
- [x] Mobile - Hero Section - Left-align CTAs
- [x] Mobile - Reduce font size of "Trusted by" (make consistent or smaller than the main heading)
- [x] Desktop - Add spacing between GIF and text (see screenshots)

### ✅ Acceptance Criteria 
- [x] UI issues fixed on both desktop and mobile
- [x] Checked by Betty

### 📸 Screenshots 
![image](https://github.com/tinacms/tina.io/assets/33026270/50a3a2c5-8aa8-486d-8203-dd8b8fcb1097)
**Figure: Mobile - Hero Section**

![image](https://github.com/tinacms/tina.io/assets/33026270/629541ac-0da5-4ba9-af8e-2277bcad99e6)
**Figure: Mobile - Trusted By**

![image](https://github.com/tinacms/tina.io/assets/33026270/9639c055-a676-4175-b2a4-0b977e5d27f5)
![image](https://github.com/tinacms/tina.io/assets/33026270/a85d09c9-b827-46f2-8d9d-df83401bed2e)
![image](https://github.com/tinacms/tina.io/assets/33026270/af3441ed-b6cc-4a32-aa05-942d2a93abc8)
![image](https://github.com/tinacms/tina.io/assets/33026270/b03b6433-5482-4543-9275-faf22c5804c8)
**Figure: Spacing issues**

Cc: @bradystroud @joshbermanssw

joshbermanssw:
✅ Done - checked by Betty and live

PR #1856, opened 2024-06-17
💄Fix UI Issues

joshbermanssw:
As per #1840 there were a number of formatting issues with different sized views that @bettybondoc wanted changed.

This PR addresses this - ✅ _all test passed by betty_

### DIFF
--- a/components/blocks/Features.tsx
+++ b/components/blocks/Features.tsx
@@ -7,12 +7,9 @@ import styled from 'styled-components';
 import playImage from '../../public/img/playButton.png';
 
 export function FeatureBlock({ data, index }) {
-
   const isReversed = data.isReversed;
   const isBackgroundEnabled = data.imageBackground;
-
   const isVideo = data.media && data.media[0] && data.media[0].src;
-
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


components/blocks/Features.tsx
```diff
@@ -7,37 +7,39 @@ import styled from 'styled-components';
...
-          >
-            {data.text}
-          </p>
-          <div className="flex lg:flex-row justify-evenly lg:justify-start">
+          <div className=''>
+          <div className="hidden sm:hidden lg:block lg:ml-0 lg:pl-0 lg:pb-6">
+            <hr className="!my-0 w-full" />
+          </div>
+
+            <p
```
bradystroud, line 35: Why is there no indenting here?

Do you have a formatter e.g. prettier

https://www.ssw.com.au/rules/consistent-code-style/

components/blocks/Features.tsx
```diff
@@ -7,37 +7,39 @@ import styled from 'styled-components';
...
           )}
-          <hr className="!my-0" />
-          <p
-            className="text-lg lg:text-xl lg:leading-normal block bg-gradient-to-br from-blue-700 via-blue-900 to-blue-1000 bg-clip-text text-transparent -mb-2 max-w-60ch text-balance"
-            data-tina-field={tinaField(data, 'text')}
-          >
-            {data.text}
-          </p>
-          <div className="flex lg:flex-row justify-evenly lg:justify-start">
+          <div className=''>
```
bradystroud, line 30: Is this div required if there is no classname?

bradystroud (changes requested):

components/blocks/Features.tsx
joshbermanssw: removed

components/blocks/Features.tsx
joshbermanssw: i forgot to format

components/blocks/Features.tsx
```diff
@@ -1,56 +1,65 @@
...
-import DocsRichText from '../styles/DocsRichText';
-import styled from 'styled-components';
-import playImage from '../../public/img/playButton.png';
+import { Actions } from './Actions'
+import { Container } from './Container'
+import { Prism } from '../styles/Prism'
+import { tinaField } from 'tinacms/dist/react'
+import DocsRichText from '../styles/DocsRichText'
+import styled from 'styled-components'
+import playImage from '../../public/img/playButton.png'
```
bradystroud, line 7: Not a fan of removing all the ; semicolons... I guess thats a formatting config option.

Is this consistent with the rest of the project?
I think we should make it consistent with the SSW.Website project (new pbi)

bradystroud (approved):

PR #1864, opened 2024-06-24
✨ Events Component

joshbermanssw:
As per email from @adamcogan 

> 1. After “Trusted By” propose a design for “Events” and get it checked by someone CCed
> 
> 2. You can get the next events from NDC’s site eg. NDC Porto
> 
> 3. Include past events eg. NDC Oslo
> 
> 4. Compare with other sites eg. DevExpress, Progress and Infragistics 
> 
> 5. Implement and get it checked by xx
> 

I have created a custom 3d design of our tina llama and a low poly earth.

https://github.com/tinacms/tina.io/assets/137844305/5af80c77-ed73-482b-9507-c25305885753

**Figure: Events Component**


components/blocks/VerticalCards.template.ts
```diff
@@ -0,0 +1,33 @@
...
+      ui: {
+        itemProps: (item) => ({
+          key: item.id,
+          label: item.headline,
+        }),
+      },
+      fields: [
+        { name: 'headline', label: 'Headline', type: 'string' },
+        { name: 'description', label: 'Description', type: 'string' },
+        { name: 'timeDate', label: 'Time and Date', type: 'string' },
```
bradystroud, line 25: Shouldn't this field be a `datetime` rather than a `string`?

components/blocks/VerticalCards.template.ts
```diff
@@ -0,0 +1,33 @@
...
+      list: true,
+      ui: {
+        itemProps: (item) => ({
+          key: item.id,
+          label: item.headline,
+        }),
+      },
+      fields: [
+        { name: 'headline', label: 'Headline', type: 'string' },
+        { name: 'description', label: 'Description', type: 'string' },
```
bradystroud, line 24: if there is too many characters in these fields it will looks bad on the UI :( 
You can avoid this by adding some validation 
https://tina.io/docs/extending-tina/validation/

components/blocks/VerticalCards.template.ts
```diff
@@ -0,0 +1,33 @@
...
+        { name: 'timeDate', label: 'Time and Date', type: 'string' },
+        { name: 'location', label: 'Location', type: 'string' },
+        { name: 'price', label: 'Price', type: 'string' },
```
bradystroud, line 27: Consider making this a number not a string

I dont really see the value in having the price... I'd prefer to remove it.

components/blocks/VerticalCards.template.ts
```diff
@@ -0,0 +1,33 @@
...
+        { name: 'image', label: 'Image', type: 'string' },
```
bradystroud, line 28: Consider making this a image not a string

components/blocks/VerticalCards.tsx
```diff
@@ -0,0 +1,63 @@
+import Link from 'next/link';
+import React from 'react';
+import css from 'styled-jsx/css';
+import Globe from './zzGlobe'; // Import the Globe component
+
+const Card = ({ cardItem }) => (
+  <Link href={cardItem.link || "#"}>
+    <div className="relative p-6 m-4 rounded-md group flex flex-col lg:flex-row bg-gradient-to-br from-white/25 via-white/50 to-white/75 break-inside-avoid rounded-xl shadow-md transform transition-transform duration-300 hover:scale-105 transform-origin-center">
+      <div className="w-full lg:w-1/3 lg:h-full mb-4 lg:mb-0 mr-0 lg:mr-6">
+        <img src={cardItem.image} alt={cardItem.headline} className="object-cover w-full h-full" />
```
bradystroud, line 10: Use Next/Image

components/blocks/VerticalCards.tsx
```diff
@@ -0,0 +1,63 @@
...
+      </div>
+      <div className="flex-grow flex flex-col">
+        <h2 className="font-bold text-3xl mb-2">{cardItem.headline}</h2>
```
bradystroud, line 13: Looking at the headings hierarchy, I think this should be a h3

bradystroud (commented):
Realized this was a draft PR halfway through reviewing... I thought I'd leave the comments here anyway

components/blocks/VerticalCards.tsx
```diff
@@ -0,0 +1,63 @@
+import Link from 'next/link';
+import React from 'react';
+import css from 'styled-jsx/css';
```
JackDevAU, line 3: I think we should stop using this package. 

Can you use `StyledComponents` or Tailwind instead

components/blocks/VerticalCards.tsx
```diff
@@ -0,0 +1,63 @@
...
+import Globe from './zzGlobe'; // Import the Globe component
```
JackDevAU, line 4: Should this Globe be called `zzGlobe`?

components/blocks/VerticalCards.template.ts
joshbermanssw: no because dateTime only supports a singular date, we want to be able to have the option to have from x -> y dates or just x date

components/blocks/VerticalCards.template.ts
joshbermanssw: removed description as per designers wants anywhomst

components/blocks/VerticalCards.template.ts
joshbermanssw: removed price

components/blocks/VerticalCards.tsx
joshbermanssw: ✅

components/blocks/VerticalCards.tsx
joshbermanssw: ✅

components/blocks/VerticalCards.tsx
joshbermanssw: was still WIP at this stage

components/blocks/VerticalCards.tsx
joshbermanssw: ✅

components/ui/Globe.tsx
bradystroud, line 1: Needs to handle nulls so you can add with Tina

image

bradystroud (changes requested):

components/ui/Globe.tsx
```diff
@@ -0,0 +1,196 @@
...
+        activePosition.y < 0 ? targetPosition[0] = targetPosition[0] - 30 : targetPosition[0] = targetPosition[0] - 5
+
+        const targetGeometric = new THREE.Vector3(...geographicToCartesian(targetPosition[0], targetPosition[1], 3));
+        const distance = targetGeometric.distanceTo(center);
+
+        // if (distance < restrictedRadius) {
+        //   const direction = targetGeometric.clone().normalize();
+        //   //targetGeometric.copy(direction.multiplyScalar(restrictedRadius));
+        // }
+
```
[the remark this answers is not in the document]
joshbermanssw: done

components/ui/Globe.tsx
joshbermanssw: fixed - please re-test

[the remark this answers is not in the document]
joshbermanssw: 1. Done 
2. ❌

bradystroud (approved):
Looks pretty good, as per our many conversations, there is still improvements to be made but this is great so far 🚀🦙

Issue #1879, opened 2024-07-03
🐛 Home page - Book meeting buttons inconsistent


bradystroud:
Hey @joshbermanssw 

I think these buttons should do the same thing.

Right now the Book a Demo button give you 3 options
- Jean in Europe
- Levi in Aus
- Scott in US

The Schedule Call button goes straight to Seths booking control.

- [x] Make the behaviour consistent

<img width="1451" alt="image" src="https://github.com/tinacms/tina.io/assets/38869720/7b83ca9a-e627-4d1c-a185-2ad4c6a13f8f">

PR #1890, opened 2024-07-04
Added new type of button: ModalButton - opens specified Modal

joshbermanssw:
In https://github.com/tinacms/tina.io/issues/1879 we wanted to ensure that buttons on the home-page COULD act the same as the buttons in the navigation bar - in this case we wanted <Action/> component to be able to open modals as well as open links.

This PR modifies the actions template allowing users to pick from three options: [ModalButton, LinkButton, CodeButton] in the new ButtonType field. There is also an option selector for the available modals which a user can chose from 

------------------

**Update:**

After a conversation with @bradystroud for future proofing the UX, we have created 3 seperate components for each of the buttons. Now every single time we referenced the actions component we instead get the option of different buttons to click/ use. 
This being the actions button, modal button or code button. This can easily be added to with future types of buttons. 

I have gone through all other pages and components and fixed where necessary.  

Screenshot 2024-07-05 at 9 42 02 AM

**Figure: ✨OUTDATED Modifications to Actions Template**

Screenshot 2024-07-05 at 5 44 38 PM

**Figure: 💄 Buttons Section**

Screenshot 2024-07-05 at 5 45 06 PM

**Figure: 💄Button Options**


bradystroud (approved):
LGTM

PR #1893, opened 2024-07-05
Docs: Explain `static` property for media configuration

HosamSenosi:
This pull request updates the TinaCMS documentation to clarify the behavior of the static property within the media configuration object in config.ts.

A new section has been added to explain the purpose and impact of the static property.
The explanation details how static: true makes media files static assets (uneditable through TinaCMS) and static: false (default) allows editors to upload, edit, and delete media directly within the editor.


bradystroud (approved):
LGTM thanks @HosamSenosi 🚀

Issue #1887, opened 2024-07-04
💄  Press page - move from 'products' to 'resources'


sethdaily:
(checked by Marcus)

### 🩸 Pain
It doesn't make sense to be under Products - it's not a product. It belongs with resources.

### 💼 Tasks 

- [x] Footer - move the link to 'Resources'
- [x] consider renaming 'Press' to 'Media' because it makes more sense that's where to find the logos/backgrounds

### 📸 Screenshots 
![image](https://github.com/tinacms/tina.io/assets/130118701/487ad671-d7f5-48ff-ac77-c094eee373a1)
Figure: Unclear this is where to find the logos...plus it's not a product

Cc: @marcusirmscher @landonmaxwell

PR #1896, opened 2024-07-08
💄Change "Press" to "Media"

joshbermanssw:
As per #1887 
- [ ] Change from /press to /media
- [ ] Update position on Footer from Product Item to Resource Item 

Screenshot 2024-07-05 at 12 10 00 PM

**Figure: 💄Updated Footer**


components/layout/Footer.tsx
```diff
@@ -58,6 +58,10 @@ const footerNav = [
         label: 'Support',
         link: '/docs/support',
       },
+      {
+        link: '/press',
```
bradystroud, line 62: Shouldn't this be `/media` as well?

```suggestion
        link: '/media',
```

bradystroud (changes requested):

components/layout/Footer.tsx
joshbermanssw: resolved

bradystroud (approved):

Issue #1826, opened 2024-06-04
💸 Tech debt - Update Next <Link> usages 


bradystroud:
Next Link no longer needs an <a>

1. Read the docs
https://nextjs.org/docs/pages/api-reference/components/link#legacybehavior
2. Update all usages of `<Link>`

joshbermanssw:
updated the components to get rid of child anchor tags and remove `legacyBehaviour` tags

PR #1902, opened 2024-07-10
💸 Remove 'legacyBehaviour' tag from <Link> 

joshbermanssw:
As per #1826 we have updated the <Link> components to get rid of child anchor tags and remove `legacyBehaviour` tags


bradystroud (approved):
LGTM

Issue #1873, opened 2024-07-01
💸 Tech Debt - Consistency with Images


joshbermanssw:
🩸 Pain
Sometimes for images and media we use the source location and sometimes we use the tina image property

🔎 Solution
Ensure consistency, go through all templates and ensure they use the Tina media field (within reason)

💼 Tasks
- [x] Work through templates and ensure each time media is used we use Tina media field and not the src

joshbermanssw:
✅ Done

PR #1905, opened 2024-07-15
💸 Tech Debt - Img Src using Media Manager

joshbermanssw:
As per #1873 I previously just used a string for an img src for components i built. The issue with this is that it limits others abilities to modify images on the site. With access to the Media Manager in Cloudinary now i can change all Image fields to type: 'Image' not type: 'String'


bradystroud (approved):
LGTM

Issue #1842, opened 2024-06-12
🤖 Dependabot - PRs fail due to no secret access 


bradystroud:
We had this problem on the SSW Website https://github.com/SSWConsulting/SSW.Website/issues/171

This makes our PR list long and ugly with lots of failures :(

- [ ] Fix

joshbermanssw:
As of right now blocked. This morning i fixed the credentials issues by providing secrets that were stored elsewhere to the dependabot. Just jumped out of a call with @ncn-ssw and we unraveled a series of bugs. 

The current error is `Error: The remote GraphQL schema does not exist. Check indexing for this branch.` even after freshly re-indexing on tina cloud. Leaving with Nick C to investigate indexing issues.

Marxoz:
Once TinaCloud has fixed this on their side, we need to do more testing to ensure it fixes all problems on Tina.io side

TinaCloud PBI: https://github.com/tinacms/tinacloud/issues/2097

PR #1906, opened 2024-07-15
🤖 Dependabot.yml Add

joshbermanssw:
As per https://github.com/tinacms/tina.io/issues/1842 we have created a dependabot.yml file so that we can manage our dependency updates better.


bradystroud (approved):
LGTM

PR #1921, opened 2024-07-16
Add recommended extensions

bradystroud:
These extensions make the DX awesome. Also issues are easily missed without having these installed


ncn-ssw (approved):
🦙

Issue #1922, opened 2024-07-17
🍪 Configure Cookie Banner to GTM Containers


joshbermanssw:
Previously we created a Cookie Consent Banner that was only hooked up to one GTM container. Now that we have access to both containers we have merged them into one, removing references to the old container. Then we need to re-wire the consent settings so that ALL tags are working with it. 

Tasks
- [x] Remove references to old container
- [x] Wire up all tags to consent settings
- [x] Test thoroughly to ensure data doesnt crash

joshbermanssw:
![Screenshot 2024-07-17 at 11 45 37 AM](https://github.com/user-attachments/assets/a1d77e3a-8b79-4e53-ae37-d4d034ebda5c)
**Figure: ✅ Merged containers, removed reference to old container and we have see there is only one GTM container at the top, and all of the correct tags are firing**

PR #1923, opened 2024-07-17
Removed references to the old GTM

joshbermanssw:
cc: @landonmaxwell 

Now that Scott has successfully exported all of the items in his GTM container, and we have imported the relevant tags into the SSW GTM. So we have removed references to the old GTM container. 

related to #1922


bradystroud (approved):
LGTM

PR #1935, opened 2024-07-23
Fix RSS feed generator

bradystroud:
Basically went to a RSS tester and fixed all the problems with it

https://validator.w3.org/feed/check.cgi?url=https://tina.io/rss.xml

Also fixed some instances of the old name tinacms.org and other broken links

Boom - igore the warning, its because its on staging
image


ncn-ssw (approved):
🦙 LGTM

Issue #1929, opened 2024-07-22
💄 Tina Docs - Pagination 


joshbermanssw:
### 🩸 Pain

When a docs page does not have a forward or backward link, the pagination looks broken (see screenshot below)

### 💡 Suggested Solution

Designer approved UI for pagination field

### 💼 Tasks 

- [x] Collaborate with designer for paginaiton
- [x] Implement 

### ✅ Acceptance Criteria 

- [x] Test passed by designer 

### 📸 Screenshots 

<img width="1049" alt="Screenshot 2024-07-22 at 12 04 34 PM" src="https://github.com/user-attachments/assets/8f597b0e-624a-4b9f-8e41-cd014dfe91af">

**Figure: ❌ Pagination looks broken **

Cc: @bradystroud @bettybondoc

bettybondoc:
cc @joshbermanssw @landonmaxwell

Expected Behaviour for Pagination:

1. Make width consistent for Previous and Next blocks

2. If only one control is available (either Previous or Next), hide the unavailable block, but do not expand the remaining block to full-width
- On the first page, there should be no Previous button (do not extend Next)
- On the last page, there should be no Next button (do not extend Previous)
- Rationale: Users expect the Previous button on the left and the Next button on the right. If the Next button extends to the left when Previous is hidden, users might mistakenly click it and navigate forward unintentionally

3. Make sure pagination is present on all pages

4. Make sure pagination looks consistent across all pages (e.g., for the Product Tour, ensure the pagination is not full-width)
Note: Avoid Overlapping Elements: Ensure that no elements overlap with the pagination controls

5. Ensure the pagination follows the logical order of the table of contents (sidebar)



![Image](https://github.com/user-attachments/assets/49ccdc7b-a621-46d8-a67a-5d69e693959c)

**Figure 1: Inconsistent width for the pagination blocks**



![Image](https://github.com/user-attachments/assets/2e634759-4ce5-40da-a4a4-3038a3e780b7)

**Figure 2: Available block is extended to full-width**



![Image](https://github.com/user-attachments/assets/51b2923b-6be9-4053-99ec-c2162cf69dd8)

**Figure 3: This page does not have pagination**



![Image](https://github.com/user-attachments/assets/2570c5b6-fa78-4439-984f-885a594dcec4)

**Figure 4: Image is overlapping  with pagination**





![Image](https://github.com/user-attachments/assets/9cdba03e-1927-4735-ad30-f9fea51de5e4)



[...]

PR #1992, opened 2024-08-02
💄UI docs pagination

joshbermanssw:
As per https://github.com/tinacms/tina.io/issues/1929 this PR addresses the UI of the docs pagination

- [ ] Having only a `next` will show the next on the right side of screen (see figure) 
- [ ] having only `prev` will show prev on the left side of screen (see figure) 
- [ ] Arrow aligned with the prev/ next

Screenshot 2024-08-02 at 5 35 24 PM

**Figure: ❌ Current - NEXT is over entire page**

Screenshot 2024-08-02 at 5 36 05 PM
**Figure: ✅ Test passed by @bettybondoc - NEXT is only on the right**

### DIFF
--- a/components/ui/DocsPagination.tsx
+++ b/components/ui/DocsPagination.tsx
@@ -23,31 +23,30 @@ export function DocsPagination({ prevPage, nextPage }: PaginationProps) {
           >
             <span className="text-sm uppercase opacity-50">Previous</span>
             <h5
-              className="text-xl leading-[1.3] m-0 transition-all ease-out duration-150 text-blue-800 group-hover:text-orange-500"
+              className="text-xl leading-[1.3] m-0 transition-all ease-out duration-150 text-blue-800 group-hover:text-orange-500 flex items-center"
             >
+              <RightArrowSvg className="w-8 h-8 fill-gray-400 transition-all ease-out duration-150 rotate-180 group-hover:fill-orange-500 mr-2" />
               {prevPage.title}
             </h5>
-            <RightArrowSvg className="absolute left-3 top-1/2 transform -translate-y-1/2 w-8 fill-gray-400 transition-all ease-out duration-150 rotate-180 group-hover:fill-orange-500" />
           </a>
         </DynamicLink>
       )}
-{nextPage && nextPage.slug && (
-  <DynamicLink href={`${nextPage.slug}`} passHref>
-    <a
-      className="col-start-2 block p-4 text-right pr-14 relative transition-all group border border-gray-100"
-      style={{ backgroundColor: '#FAFAFA' }}
-    >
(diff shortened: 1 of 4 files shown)


components/ui/DocsPagination.tsx
```diff
@@ -14,109 +14,41 @@ interface PaginationProps {
...
+            style={{ backgroundColor: '#FAFAFA' }}
+          >
+            <span className="text-sm uppercase opacity-50 pl-10">Previous</span>
+            <h5
+              className="text-xl leading-[1.3] m-0 pl transition-all ease-out duration-150 text-blue-800 group-hover:text-orange-500 flex items-center"
+            >
+              <RightArrowSvg className="w-8 h-8 fill-gray-400 transition-all ease-out duration-150 rotate-180 group-hover:fill-orange-500 mr-2" />
+              {prevPage.title}
+            </h5>
+          </a>
```
bradystroud, line 31: Shouldn't we use `<Link>` here?

pages/blog/[slug].tsx
```diff
@@ -48,36 +52,47 @@ export const components: Components<{
...
+      <div className="grid grid-cols-2 gap-4">
+        <div className="bg-red">
+          {' '}
+          <TinaMarkdown
+            content={props.docText as any}
+            components={components}
+          />{' '}
+        </div>
+        <div>
+          <img src={props?.image} alt="image" className="w-full" />
```
Comment (bradystroud) [inline pages/blog/[slug].tsx:64]: Shouldn't we use Next Image here?

pages/docs/product-tour.tsx
```diff
@@ -169,7 +169,9 @@ export default function Page(props) {
               </div>
             </SplitContent>
             <LastEdited date={doc_data.last_edited} />
-              <DocsPagination prevPage={previousPage} nextPage={nextPage} />
+            <div className='w-1/2'>
+                            <DocsPagination prevPage={previousPage} nextPage={nextPage} />
+            </div>
```
bradystroud, line 174: formatting seems wrong here

bradystroud (changes requested):

pages/blog/[slug].tsx
Comment (joshbermanssw) [reply pages/blog/[slug].tsx:64]: Yes, fixed.

pages/docs/product-tour.tsx
joshbermanssw: Resolved

components/ui/DocsPagination.tsx
joshbermanssw: Usually yes, However we dont use navigation in our <a> tags, we do so in our <DynamicLink>. Rather ive replaced our <a> with <div>'s

bradystroud (approved):
LGTM

Issue #1883, opened 2024-07-03
🧠 Spike | Investigate 3D Model Performance Issues (Event Map)


joshbermanssw:
cc: @wicksipedia @bradystroud 

Investigate how we can improve the performance of our webpage with the ultimate goal of improving SEO - 

**From Matt W**
Hey Josh
 
Has this feature affected the performance score for the page?
 
Given that the component is not at the top of the page, should it’s loading be deferred?
 
![image003](https://github.com/tinacms/tina.io/assets/137844305/af830c9d-ee38-4bfc-9df2-85268524dc2d)

**Figure: The last 28 days – its all green, but the current scores are red**

joshbermanssw:
Update 7/8 **WITHOUT THE GLOBE** 

![Image](https://github.com/user-attachments/assets/54c5d730-02a4-41b5-bf66-44d608ac53ff)

**Figure: Mobile Performance**

![Image](https://github.com/user-attachments/assets/369f3eb9-4826-463d-8a68-8dbbc539f0f0)

**Figure: Desktop performance**

joshbermanssw:
cc: @landonmaxwell 

Cloud Cannon : https://pagespeed.web.dev/analysis/https-cloudcannon-com/t30649sw5w?form_factor=desktop&category=performance&category=accessibility&category=best-practices&category=seo&hl=en-US&utm_source=lh-chrome-ext

Decap: https://pagespeed.web.dev/analysis/https-decapcms-org/qt0l178jrm?form_factor=desktop&category=performance&category=accessibility&category=best-practices&category=seo&hl=en-US

joshbermanssw:
cc: @landonmaxwell 
We want to look into improving the speed of the site in general. if the globe does NOT impact the speed, re-implement. Otherwise focus on boosting the sites performance before re-implementing the globe

joshbermanssw:
![Image](https://github.com/user-attachments/assets/767e4311-fde4-4c49-9867-fa2461b5fd29)
![Image](https://github.com/user-attachments/assets/2e925a17-e267-4be1-b2c1-88a51140be92)

joshbermanssw:
**Applied Lazy Loading**

![Image](https://github.com/user-attachments/assets/8d20fab7-ce5b-4aff-933e-60b8c5ad4034)
![Image](https://github.com/user-attachments/assets/eff83623-b63a-4af8-ac52-af6a7bb3e62a)

joshbermanssw:
[...]

PR #2038, opened 2024-08-19
⚡ Optimize events component

joshbermanssw:
Apart of this sprint i investigated how we can re-introduce the events component in a way that doesnt negatively impact the performance of the tina.io site. My findings in https://github.com/tinacms/tina.io/issues/1883 saw that lazy loading actually increased the performance of the tina.io page. 

image
**Figure: Findings from spike**


bradystroud (approved):
LGTM
Seems fast ⚡

Issue #2006, opened 2024-08-07
🔨 TinaCMS Packages Audit


joshbermanssw:
cc: @bradystroud 
The packages.json file has some packages we do not need anymore - we do not need them.

- [x] Go through packages.json and remove unrequired packages

Be strict - follow this rule
https://www.ssw.com.au/rules/installing-3rd-party-libraries/#1-is-it-valuable

joshbermanssw:
**BEFORE CHANGES**

Using a module called `depcheck` 

![Image](https://github.com/user-attachments/assets/6a9e2e8d-8f5c-47f2-8e8a-035d2c3dc9bb)

using `npm outdated`

![Image](https://github.com/user-attachments/assets/3c5e58d9-97af-463d-b86c-1200e0f7c485)

joshbermanssw:
✅ Done removed unused packages 



![Image](https://github.com/user-attachments/assets/cbd59104-331f-42d0-889d-4c1e8c8c1c5d)

**Figure: Running `depcheck` before audit**

![Image](https://github.com/user-attachments/assets/a3c3d1d6-f3dc-41eb-9666-3fec3045cf22)

**Figure: Running `depcheck` after audit**

PR #2059, opened 2024-08-21
📦 Dependency Audit - How many unused dependencies can we get rid of? 

joshbermanssw:
This PR addresses PBI https://github.com/tinacms/tina.io/issues/2006 - essentially cleaning up the list of unused dependencies we have. I have used `depcheck` as the tool to identify unused dependencies. Although it is not always completely accurate i have managed to remove around 60% of the dependencies it has identified as unused. the others were identified as unused but were actually used. 

357712392-6a9e2e8d-8f5c-47f2-8e8a-035d2c3dc9bb

**Figure: running `depcheck` before**

Screenshot 2024-08-30 at 4 21 20 AM

**Figure: after making deletions and running `depcheck`**


bradystroud (changes requested):

bradystroud (approved):
LGTM - everything worked

PR #2060, opened 2024-08-21
Fix pricing sizing

joshbermanssw:
cc: @wicksipedia @bettybondoc

For the new pricing component there were issues from the lg-1250px view. So we extended the medium view up until 1250px and then from 1250px onwards uses the large view


isaaclombardssw (approved):
Tailwind would be preferred, but as we want custom breakpoints here it looks okay.

Marxoz (approved):

Issue #2071, opened 2024-08-21
💄 Pricing component | Mobile UI 


joshbermanssw:
- [x] Increase size of includes text for mobile view

joshbermanssw:
Live

PR #2072, opened 2024-08-21
💄Increase size of text for mobile

joshbermanssw:
Stemming from #2071, we've just bumped up the size of the text in the includes section for mobile view 

Screenshot 2024-08-21 at 5 27 28 PM

**Figure: Updated View**


isaaclombardssw (approved):
👍 👍

PR #2074, opened 2024-08-21
💄 Mobile and Tablet UI changes to the Events Component

joshbermanssw:
- [ ] Hides globe for small view 
- [ ] Fixed spacing 

still needs to remove clickable cards for md mode


components/blocks/Events.tsx
```diff
@@ -59,14 +59,10 @@ const Card = ({ cardItem, onHover }) => {
           )}
         </div>
         <div className="flex-grow flex flex-col pl-4">
-          <div><h3 className="font-bold text-3xl mb-1 inline">{cardItem.headline}</h3>
-            {new Date() > new Date(cardItem.endDate) ? <span className="rounded-full w-24 ml-3 inline items-center font-tuner whitespace-nowrap text-gray-800 px-4 pt-[5px] pb-[5px] text-sm font-medium border border-gray-300 bg-gradient-to-br from-white to-gray-200">
-              Done
-            </span> : <></>}
-            </div>
```
isaaclombardssw, line 66: This is removing a 'done' marker next to complete events that Adam asked for. (on the current homepage https://tina.io/). Could this be preserved?

It was also suggested that past events could be greyed out as an alternative option.

isaaclombardssw (changes requested):

isaaclombardssw (approved):

components/blocks/Events.tsx
isaaclombardssw: Josh'll handle this one later 👍

Issue #2138, opened 2024-09-02
Icon Selector - allow no icon to be selected


bradystroud:
Really smooth UI. Should there be a way to deselect icons (i.e. none selected)?

If we don't want that, it's great.

_Originally posted by @isaaclombardssw in https://github.com/tinacms/tina.io/pull/2137#pullrequestreview-2276189180_

joshbermanssw:
✅ Done and live 

![Image](https://github.com/user-attachments/assets/42cb5ba9-3f9f-461d-a6a3-200a75aae448)

PR #2149, opened 2024-09-04
✨ No Icon option for Icon Selector Component 

joshbermanssw:
As per https://github.com/tinacms/tina.io/issues/2138 i have added an extra option to the IconSelector component so that users can opt for 'No Icons' 

Screenshot 2024-09-04 at 8 03 12 PM

**Figure: No Icon option added**


isaaclombardssw (approved):
Looks great

Issue #1985, opened 2024-08-01
📃Fix broken link of documentation


KristenHu:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

This broken link in the ["Using Multiple Branches"](https://tina.io/docs/tina-cloud/branching/) page needs to be fixed.

![Image](https://github.com/user-attachments/assets/fe6db133-b6e5-44b8-b1ed-de7dbe2e701e)

### Reproduction

https://tina.io/docs/tina-cloud/branching/

### System Info

Browsers: Chrome

### Validations

- [x] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://tina.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

wicksipedia:
unable to reproduce

piers-sinclair:
I am having this issue on my machine as well


![Image](https://github.com/user-attachments/assets/99f036af-7e85-4b2f-b1e8-17b07094e77c)

**Figure: A gif of the problem**

wicksipedia:
ok, looks like the repo is private and that is why the link isn't working for you

ncn-ssw:
@wicksipedia @piers-sinclair 

The link is broken as the repository at https://github.com/tinacms/demo-incremental/tree/main is private.

wicksipedia:
Repo has been made public - link should now work

PR #2152, opened 2024-09-04
📄 Removed public reference to private repo

joshbermanssw:
We had a reference to a private repo on https://tina.io/docs/tina-cloud/branching/. The related PBI is https://github.com/tinacms/tina.io/issues/1985.

This PR removes the reference to private repo so users dont think the link is broken

### DIFF
--- a/content/docs/tina-cloud/branching.mdx
+++ b/content/docs/tina-cloud/branching.mdx
@@ -19,8 +19,6 @@ cmsCallback: (cms) => {
 ```
 
 \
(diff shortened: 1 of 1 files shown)


bradystroud:
Do you think we still need an example? Should we make a public one?

bradystroud (changes requested):
It's a demo repo... I think it makes more sense to make it public.

I'd check with @wicksipedia

wicksipedia:
No longer required - we have made the repo public

PR #1816, opened 2024-06-03
💄🏗️ Homepage - Key Features rework 

joshbermanssw:
⚠️ WIP NOT DONE - WAITING FOR MEDIA

Issue #1805, opened 2024-05-29
✨ Key Features - Homepage Carousel 


joshbermanssw:
### 🩸 Pain

The current key features boxes arent the best looking, and they are located quite far down the page. 

### 💡 Suggested Solution

Refer to the revamped homepage to implement the carousel styled key features box and move location up. 

[Figma Mockup](https://www.figma.com/proto/Jyj0SszjfSBc6Minkl3g1l/Tina.io?node-id=248-3018&t=esyxMgKmsNVIlFfr-1&scaling=scale-down&page-id=248%3A3017&starting-point-node-id=248%3A3018)

### 💼 Tasks 

- [x] Create carousel styled component for the CMS 
- [x] Ensure it is automatically scrolling vertically (look at react-slick) 
- [x] Replace current feature boxes
- [x] Add GTM tags to each individual new feature block (refer to #1802)
- [x] Implement and Test 

### ✅ Acceptance Criteria 

- [ ] Test passed by designers and product owner 

### 📸 Screenshots 

Cc:

bettybondoc:
Additional Tasks:
- [ ] Consult with Betty for the mobile view
- [ ] Get final copy/GIFs from Landon

Expected Behaviour (See https://webflow.com/)
- Desktop - Carousel will cycle through each item automatically. When a user hovers over an item, the carousel stops. Clicking on an item will open a new tab.
- Mobile - All the items will be displayed at once (no hovering).

<img width="475" alt="image" src="https://github.com/tinacms/tina.io/assets/33026270/7eeca86a-d40c-4ac7-84ca-90dc4c2a8951">

**Figure: Key Features**

joshbermanssw:
Awaiting GIFs still

Marxoz:
@landonmaxwell to supply @joshbermanssw with the GIFs

joshbermanssw:
Update - code has been approved, content not yet finished. Awaiting 2 more animations

PR #2164, opened 2024-09-06
✨ Key Features | Homepage Carousel and Cleanup

joshbermanssw:
This is a component I had built a while ago but was never finished due to needing media for it. However because it is so old and we have updated some of its dependent components/ methods i decided it'd be easier to re-build it. 

PBI https://github.com/tinacms/tina.io/issues/1805

Old PR https://github.com/tinacms/tina.io/pull/1816

This is a component built from @bettybondoc's original re-design of the homepage from figma: https://www.figma.com/design/Jyj0SszjfSBc6Minkl3g1l/Tina.io?node-id=399-5400&node-type=FRAME

**Changes:**
- added another button variant, ghost (blue) which is the same as ghost normal but blue text instead of red
- created new CarouselFeature component and template
- removed 2 old feature blocks from the homepage (as per figma design) 

**Still todo**
- [x] media rendering with using the `image` field rather than src


isaaclombardssw (changes requested):
Sick component 🚀 

Only thing that has to change is the below (missing content in 1st and 3rd RHS option).

components/blocks/CarouselFeature.tsx
```diff
@@ -0,0 +1,265 @@
...
+        intervalRef.current = setInterval(() => {
+          setHoveredIndex((prevIndex) => {
+            if (prevIndex === null || prevIndex === data?.items?.length - 1) {
+              return 0;
+            }
+            return prevIndex + 1;
+          });
+        }, 3000);
+      }
+    };
```
isaaclombardssw, line 156: Could do with some comments.

isaaclombardssw (approved):
Code looks aight 👍

components/blocks/CarouselFeature.tsx
```diff
@@ -0,0 +1,264 @@
...
+  FaFileAlt,
+  AiOutlineUser,
+  BiBadge,
+  BiSupport,
+  AiOutlineUsergroupAdd,
+  CgCrown,
+  HiOutlineSparkles,
+  TbPlugConnected,
+  SlLock,
+};
```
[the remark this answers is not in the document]
joshbermanssw: done

components/blocks/CarouselFeature.tsx
joshbermanssw: done

PR #2193, opened 2024-09-13
🔗 removing private links from whats-new/tinacloud

joshbermanssw:
We provided links to the TinaCloud repository. This repository is private. Meaning the links would be broken and 404 for external users


bradystroud:
Why?
Can you link the issue that triggered this change?

ncn-ssw:
**question**: Why is `yarn.lock` being modified?

joshbermanssw:
See https://github.com/tinacms/tina.io/pull/2199, made a new branch now that i've fixed my local

PR #2255, opened 2024-09-27
Revert "➖ Old Dependencies – Removing @tinacms/graphql-old and /datalayer-old"

isaaclombardssw:
This is kind of interesting – Jeff mentioned the existence of a Tina playground and we thought nobody had access to it anymore / was using it.

It's not something we include on the site afaik.

Removing this package seems to have broken it – can be found at https://tina-gql-playground.vercel.app/object-list-templates.


joshbermanssw (approved):
lgtm

Issue #2219, opened 2024-09-18
💄 Tina-Docs - Search colour


joshbermanssw:
cc: @landonmaxwell 

When searching in the Tina docs, all the text is orange - this makes the page busier than it needs to be

Either:
- [x] Make all of the text black and perhaps a highlight on hovered 
- [x] Make hyperlinks orange and body text black 

![Image](https://github.com/user-attachments/assets/d1465578-f0e6-4d48-938d-9f8b6e963bda)

joshbermanssw:
![Image](https://github.com/user-attachments/assets/c3cf166e-14f1-4b13-a7f8-fbe52d36da10)

**Figure: New Search Docs Colour**

![Image](https://github.com/user-attachments/assets/ed4d360d-d2eb-41ef-b85c-671c74c276d5)

**Figure: Hover - New Search Docs Colour**

PR #2295, opened 2024-10-04
💄 Reduce TinaDocs search noise 

joshbermanssw:
As per https://github.com/tinacms/tina.io/issues/2219 we wanted to reduce the noise in the search for users in our docs 

**Changes**
- Kept underlined hyperlinks orange
- Made body text the dark navy to match the rest of docs
- On hover body will be same colour as hyperlink 

Screenshot 2024-10-04 at 4 23 58 pm
**Before**

Screenshot 2024-10-04 at 4 21 59 pm
**After - and hovering on first option**


isaaclombardssw (approved):
Looks good – obligatory reminder styled components are total tech debt for us.

Issue #2214, opened 2024-09-17
⬆ Updating Dependencies – Remark


isaaclombardssw:
### Pain

I was attempting to update Remark (dependency) and ran into some issues since it's become ES module only.

### Tasks

- [ ] Update Remark to a later version (15+) and debug transpilation errors.

joshbermanssw:
We were using the Remark dependency to call one method from it that was only used once. The purpose of this method was to strip away the markdown content so that we could better index our content that was type `TinaMarkdownContent` to be used in our Algolia searching for Docs and Blogs.

I removed the strip-markdown and remark dependencies and replaced them with a custom markdown strip function (essentially alot of regex stripping) and added some test-cases to test for edge cases.

So instead of updating the dependency as per task spec, i removed it. This is a better outcome as its one less dependency we need to maintain the versioning of.

PR #2314, opened 2024-10-09
remove <a/> from passing into Link

joshbermanssw:
Whilst attempting to upgrade the remark dependency i found a legacyBehaviour issue with anchor tags being passed into the `DynamicLink` component. So i've removed the anchors and replaced them with span tags 

[⬆ Updating Dependencies – Remark](https://github.com/tinacms/tina.io/issues/2214)


isaaclombardssw (approved):
Good catch 🙏

PR #2316, opened 2024-10-09
👥 remove strip markdown and remark

joshbermanssw:
So we had a PR https://github.com/tinacms/tina.io/issues/2214 to update a dependency `remark`

However i noticed this one dependency was only being used once througout the entire code base for a singular method. Specifically it was beign used to remove markdown syntax from our .md and .mdx files so that we can index them to use in our algolia searches.

In this pr i : 
- removed the strip-markdown and remark dependencies 
- removed indexing on guides as we moved them into docs
- re-wrote the `stripMarkdown` method in our `blog_helper.ts` modelled after https://github.com/zuchka/remove-markdown
- added a test-case to ensure our updated method works for an edge case i found whilst building the new function 

Screenshot 2024-10-09 at 8 55 28 am

**Figure: ✅ Algolia powered Doc search working**

Screenshot 2024-10-09 at 5 38 13 pm

**Figure: ✅ All test cases passing!**


utils/blog_helpers.ts
```diff
@@ -7,25 +7,44 @@ export function orderPosts(posts) {
...
+      .replace(/!\[([^\]]*)\]\([^\)]+\)/g, '$1')
+      .replace(/#{1,6}\s*(.*)/g, '$1')
+      .replace(/>\s*(.*)/g, '$1')
+      .replace(/(\*\*|__)(.*?)\1/g, '$2')
+      .replace(/(\*|_)(.*?)\1/g, '$2')
+      .replace(/`([^`]+)`/g, '$1')
+      .replace(/^(\n)?\s{0,}#{1,6}\s*( (.+))? +#+$|^(\n)?\s{0,}#{1,6}\s*( (.+))?$/gm, '$1$3$4$6')
+      .replace(/~(.*?)~/g, '$1');
+
+    content = content.replace(/<[^>]*>/g, '');
```
isaaclombardssw, line 33: AI thinks this is alright.

🤖
The code is not "weird," but it is quite aggressive in sanitizing and formatting text. It aims to strip out various types of formatting, tags, and lists to convert the content to plain text.
🤖

isaaclombardssw (approved):

utils/blog_helpers.ts
bradystroud: FYI, we checked out a few packages that achieve the same results and they use very similar code although some more comments could be useful @joshbermanssw  😉

bradystroud: Also the tests help with the code being a bit scary.

diff.txt
bradystroud, line 1: 🚮

bradystroud (changes requested):

utils/blog_helpers.ts
joshbermanssw: added some more comments! Did you give AI the context that the method is used as a markdownStripper?

diff.txt
joshbermanssw: ✅

bradystroud (approved):

PR #2399, opened 2024-10-31
SEO & Redirects - Remove Trailing Slash and Fix Redirects

joshbermanssw:
This PR is to address [#24](https://github.com/tinacms/business-backlog/issues/24)

Essentially we were seeing in our SEMRush audits that we had a series of long redirect chains. However they were primarily because we were using the 'trailingSlash: true' which saw redirects that did not have an appended final '/' having added them.

i.e if we re-routed a user to `/legacy-redirect`, our config was then re-routing to `/legacy-redirect/`

Adding this extra forward slash was increasing this chain. 

@landonmaxwell tested this branch on SEMRush and found that this fixed the issue o redirect chains and loops

image (7)
**Figure: Long Redirect chains eliminated**

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -86,7 +86,6 @@ const config = {
       },
     ]
   },
-  trailingSlash: true,
   exportPathMap: async function () {
     return {}
   },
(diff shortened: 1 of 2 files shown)


isaaclombardssw (approved):
Small change big impact 👍

Issue #2403, opened 2024-11-01
🐛  Docs ToC | Once again, not highlighting the current page


isaaclombardssw:
### 🐛 Describe the Bug
The current page is not being highlighted on the docs ToC.

### 🧑‍💻 Reproduction Steps

1. Go to https://tina.io/docs/product-tour
2. Look at the ToC (should be able to see Introduction | Introduction To TinaCMS highlighted)
3. See error 
Note: many pages effected

### 🎯 Expected Behaviour 
Highlight

### 💼 Tasks
- [ ] Investigate and fix

### ✅ Acceptance Criteria 
All ToC items highlight when on the relevant page.

joshbermanssw:
I investigated as to why the highlighting wasnt working - through a series of logs i found it was because one of the conditions of a nav item being highlighted was !categoryData.items && !selected.

Upon further looking it turns out that the selected boolean wasnt being set correctly due to a trailing back slash.

This would be due to the removal of automatic trailing backslashes made last week as we aimed to reduce our redirect chains (see https://github.com/tinacms/tina.io/pull/2399 for more context)

This PR just removes the trailing backslash at the end of the slug var using regex

joshbermanssw:
![Image](https://github.com/user-attachments/assets/c5d90937-2383-44cd-8215-d44b5d8fad08)

**Figure: Docs highlighting on live site**

PR #2441, opened 2024-11-08
🌠 Fix docs highlighting 

joshbermanssw:
As per https://github.com/tinacms/tina.io/issues/2403 

I investigated as to why the highlighting wasnt working - through a series of logs i found it was because one of the conditions of a nav item being highlighted was `!categoryData.items && !selected`. 

Upon further looking it turns out that the `selected` boolean wasnt being set correctly due to a trailing back slash. 

This would be due to the removal of automatic trailing backslashes made last week as we aimed to reduce our redirect chains (see https://github.com/tinacms/tina.io/pull/2399 for more context) 

This PR just removes the trailing backslash at the end of the `slug` var using regex

### DIFF
--- a/components/DocumentationNavigation/DocsNavigationList.tsx
+++ b/components/DocumentationNavigation/DocsNavigationList.tsx
@@ -17,6 +17,7 @@ interface NavTitleProps {
   onClick?: () => void;
 }
 
+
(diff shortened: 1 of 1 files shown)


isaaclombardssw (approved):
LGTM

Issue #2251, opened 2024-09-26
♻ Refactor – Remove `Guides` stuff


isaaclombardssw:
### Pain

We've removed the Guides part of the website into the docs, but there still exists guide-specific code.

### Solution

Try and nuke the code relating to guides.

### Tasks

- [x] Delete `utils/guide_helpers.ts`
- [x] Delete `utils/fetchGuides.ts`
- [x] Delete `components/DocumentNavigation/DocsHeaderNav.tsx`
- [x] Test

Cheers 🦙

joshbermanssw:
Done and live

PR #2498, opened 2024-11-20
🗑️ Remove Old Guides Content

joshbermanssw:
As per #2251 

This PR just removes all references left to `Guides` that arent apart of our new /docs/guides

It also updates the link on the 404 page from /guides to /docs/guides


isaaclombardssw (approved):
LGTM

PR #2516, opened 2024-11-25
Update packages and remove styled-jsx-plugin-sass

bradystroud:
as per https://github.com/tinacms/tina.io/issues/2517 

Pair programming with @joshbermanssw fixing the node-sass build issue

We didn't really clean up any tech debt here, we should do that ASAP though


joshbermanssw (approved):
lgtm

Issue #2500, opened 2024-11-20
💄 Comparison Box – Images to use Image Type


isaaclombardssw:
### 🩸 Pain
Currently the template for the comparison box takes in a link: 

![Image](https://github.com/user-attachments/assets/a54d353c-1bee-4808-a85d-ee2201405c52)

Tina has a great image type we could be use instead.

### 💼 Tasks 

- [ ] Change to image and upload all the current logos.

joshbermanssw:
Fixed and live

PR #2559, opened 2024-11-29
Change string to Image in CompareBox schema

joshbermanssw:
As per https://github.com/tinacms/tina.io/issues/2500 I have changed the schema to use the image field rather than string. Not a big change because the image field just grabs the url when pushing the data to the component


isaaclombardssw (approved):
great change

PR #2565, opened 2024-12-01
Kd/fix/broken self hosted faq link

joshbermanssw:
fixing broken link


ncn-ssw (approved):
🦙 Link checked correct

Issue #2613, opened 2024-12-16
☄️ Tina.io - install Clarity


landonmaxwell:
For website insights, install Microsoft Clarity.
https://clarity.microsoft.com/

Note: We have HotJar already installed, but want to compare results for both and choose which is better.

joshbermanssw:
Done and live ✅

PR #2619, opened 2024-12-16
🌠 Add Microsoft Clarity

joshbermanssw:
cc: @landonmaxwell 

As per #2613 wwe have installed microsoft clarity

### DIFF
--- a/pages/_document.tsx
+++ b/pages/_document.tsx
@@ -45,18 +45,6 @@ export default class MyDocument extends Document {
               `,
             }}
           />
-          {/* Microsoft Clarity */}
-          <script
-            dangerouslySetInnerHTML={{
-              __html: `
-                (function(c,l,a,r,i,t,y){
-                  c[a]=c[a]||function(){(c[a].q=c[a].q||[]).push(arguments)};
-                  t=l.createElement(r);t.async=1;t.src="https://www.clarity.ms/tag/"+i;
-                  y=l.getElementsByTagName(r)[0];y.parentNode.insertBefore(t,y);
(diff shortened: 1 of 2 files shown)


pages/_document.tsx
```diff
@@ -44,6 +45,18 @@ export default class MyDocument extends Document {
...
+            dangerouslySetInnerHTML={{
+              __html: `
+                (function(c,l,a,r,i,t,y){
+                  c[a]=c[a]||function(){(c[a].q=c[a].q||[]).push(arguments)};
+                  t=l.createElement(r);t.async=1;t.src="https://www.clarity.ms/tag/"+i;
+                  y=l.getElementsByTagName(r)[0];y.parentNode.insertBefore(t,y);
+                })(window, document, "clarity", "script", "pepjushhm5");
+              `,
+            }}
+          />
```
isaaclombardssw, line 59: Wonder if we can do it like with Hotjar – https://github.com/tinacms/tina.io/blob/main/pages/_app.tsx (see the useEffect)

isaaclombardssw (changes requested):

pages/_document.tsx
isaaclombardssw: Also should we remove Hotjar if we're moving to MS clarity.

isaaclombardssw (approved):
LGTM

pages/_document.tsx
joshbermanssw: Done - we're keeping both HotJar and Clarity to compare coverage for now

Issue #2628, opened 2024-12-17
🚏 Migrate Root Layout (_document and _app) files 


joshbermanssw:
Starting step for moving from pages to app router - see [Step 2: Creating a Root Layout](https://nextjs.org/docs/pages/building-your-application/upgrading/app-router-migration#step-2-creating-a-root-layout) 

Replacing _app.tsx and _document.tsx with layout.tsx

joshbermanssw:
✅ DOne and Live

PR #2646, opened 2024-12-20
🚏 Pages to App Router Initiation

joshbermanssw:
As per [Migrate Root Layout (_document and _app) files - as per next js docs this needs to be done first](https://github.com/tinacms/tina.io/issues/2628) this PR accounts for moving over the root layout to start the pages to app router migration. 

To test how it works and add a route I also migrated the /events route over to the app router

_Note the PR has a lot of files changes BECAUSE of the duplicated components used as client components in app router and also used in current pages router pages_

**Preview: https://tina-io-git-page-to-app-tinacms.vercel.app/?vercelToolbarCode=GSoynmcoJ3z7If4#**

**Changes**
- [ ] Created a `/componnets/AppRouterMigrationComponents/` directory to hold all the high level components used in `layout.tsx` including new `SiteLayout.tsx` 
- [ ] Moved the styled.jsx which styled the sites background from `Layout.tsx` (old layout file) to the tailwind.config as `blob-bg` and used in both `SiteLayout.tsx` and `Layout.tsx` (both layout files) 
- [ ] Moved /events route to app router, with associated `/events/page.tsx` and `/events/EventsClient.tsx` 
- [ ] All components created in `/componnets/AppRouterMigrationComponents/` are client duplicates of their page router counterparts


app/events/page.tsx
```diff
@@ -0,0 +1,12 @@
+// app/events/page.tsx
+import { client } from 'tina/__generated__/client';
+import EventsClient from './EventsClient';
+
+export default async function EventsPage() {
+  const vars = {};
+
+  const res = await client.queries.eventsConnection(vars);
```
babakamyljanovssw, line 8: The `getStaticProps` functions which was used in old Pages Router is handled directly in `pages.tsx` and passed to client component as props when migrating to App Router.
Please read this documentation for more info - https://nextjs.org/docs/app/building-your-application/upgrading/app-router-migration#static-site-generation-getstaticprops

For example:
```
const getEvents = async () => {
  const res = await client.queries.eventsConnection(vars);
  return {
    props: {
      data: res.data
      data: res.query
    }
  }
}
```

And then when calling the component from `page.tsx`:
```
export default async function EventsPage() {
  const { props } = await getEvents();

  return <EventsClient props={props} />;
}
```

pages/events.tsx
```diff
@@ -1,159 +0,0 @@
...
-    const cards = document.querySelectorAll('.event-card');
-    cards.forEach((card) => observer.observe(card));
-
-    return () => observer.disconnect();
-  }, [visibleCards]);
-
-  return (
-    
-    <Layout>
-      <DefaultSeo
```
babakamyljanovssw, line 87: I couldn't find where did you configured seo in App Router.
This can be handled using `generateMetadata` function - https://nextjs.org/docs/app/building-your-application/optimizing/metadata.

Add `generateMetadata` to `page.tsx` file in App Router to set seo fields

app/events/page.tsx
joshbermanssw: Done

app/events/page.tsx
```diff
@@ -0,0 +1,18 @@
...
+import EventsClient from './EventsClient';
+
+export default async function EventsPage() {
+  const vars = {};
+
+  const { data, query } = await fetchEvents(vars);
+
+  return <EventsClient query={query} data={data} vars={vars} />;
+}
+
```
babakamyljanovssw, line 11: Please add generateMetadata here too

joshbermanssw: DOne

babakamyljanovssw (approved):
LGTM 🚀

components/AppRouterMigrationComponents/AdminLink.tsx
```diff
@@ -0,0 +1,43 @@
...
+      {showAdminLink && (
+        <div className="fixed top-4 right-4 flex items-center justify-between bg-blue-500 text-white px-3 py-1 rounded-full z-50">
+          <a
+            href={`/admin/index.html#/~${window.location.pathname}`}
+            className="text-xs"
+          >
+            Edit This Page
+          </a>
+          <button onClick={handleDismiss} className="ml-2 text-sm">
+            xx
```
wicksipedia, line 35: time to fix this bug
can you use a cross icon instead

components/AppRouterMigrationComponents/AppFooter.tsx
```diff
@@ -0,0 +1,324 @@
...
+import { BsDiscord } from 'react-icons/bs';
+import GithubIconSvg from '../../public/svg/github-icon.svg';
+import LinkedInIconSvg from '../../public/svg/linkedin-icon.svg';
+import XIconSvg from '../../public/svg/x-icon.svg';
+import YoutubeIconSvg from '../../public/svg/youtube-icon.svg';
+import '../../styles/tailwind.css';
+import { TinaIcon } from '../logo';
+import { DynamicLink } from '../ui';
+
+const footerNav = [
```
wicksipedia, line 14: is there a PBI to migrate this hardcoded content into TinaCMS?
add a comment with a link to it here

components/AppRouterMigrationComponents/CloudBanner.tsx
```diff
@@ -0,0 +1,49 @@
...
+          <a
+            href="https://github.com/tinacms/tinacms"
+            target="_blank"
+            rel="noopener noreferrer"
+            className="underline"
+          >
+            GitHub
+          </a>
+        </span>
+        {/* @ts-ignore */}
```
wicksipedia, line 37: why ts-ignore?

wicksipedia:
why are the new components under `components/AppRouterMigrationComponents`?
I'm guessing that these are the components used on the new app router views, maybe add some documentation as to why

wicksipedia:
The events page seems to render a few things differently as well, not sure if these are expected:
- navbar has different letter spacing resulting in different sized buttons and elements rendering in different locations
- event page - different font is being used for header text

CleanShot 2024-12-23 at 15 11 10

components/AppRouterMigrationComponents/AdminLink.tsx
joshbermanssw: ✅ Done

components/AppRouterMigrationComponents/AppFooter.tsx
joshbermanssw: Done, commented in both locations

components/AppRouterMigrationComponents/CloudBanner.tsx
joshbermanssw: Removed

joshbermanssw: Removed

wicksipedia (approved):
lgtm

Issue #2668, opened 2024-12-29
📝 Do not display default Last Edited date


AntPolkanov:
### 💫 Describe the Issue With the Current Docs

Pages have "Last Edited" date which is set to a default value January 1, 1970 (see Figure 1 below)

![Image](https://github.com/user-attachments/assets/6a6c3db6-dc6d-4f07-b6ab-fd3d995fb7db)

**Figure 1**

### 📝 Outline the Proposed Change

1. Hide Last Edited label if the value is equal to default.

### 💼 Tasks

- [x] Fix it

### 📸 Screenshots 

Cc:

joshbermanssw:
Fixed and live

PR #2681, opened 2025-01-03
📑 docs - hide date if it is null

joshbermanssw:
As per #2668 some of the docs pages that havent been updated were displaying a last edited date of January 1970. This is because the date was null and it was being parsed into a default date with `Date`.

To combat this i've hidden dates if they are null


isaaclombardssw (approved):
LGTM

Issue #2553, opened 2024-11-29
Move /whats-new routes to the app router


joshbermanssw:

joshbermanssw:
Done and live

PR #2690, opened 2025-01-06
🚏 Move /whats-new pages to app router

joshbermanssw:
As per #2553 i moved the whats-new pages to the app router, with the same pattern as other files. Very simple change as there are no client side components used on these pages


isaaclombardssw (approved):
LGTM ✅

PR #2718, opened 2025-01-13
next.config.js - Add security headers

wicksipedia:
Adds security headers to next config to prevent loading the site in an iframe

**Before:**
CleanShot 2025-01-13 at 12 24 09@2x

**After:**
CleanShot 2025-01-13 at 12 24 17@2x


ncn-ssw (approved):
🦙 LGTM

PR #2721, opened 2025-01-13
next.config.js - Update security headers for frame options and content security policy

wicksipedia:
This should allow the site to be editable in the Tina editor (since it needs to load the site in an iframe on the same origin)


ncn-ssw (approved):
🦙 LGTM

Issue #2695, opened 2025-01-07
💄 Footer - Mega Menu Improvement


joshbermanssw:
cc: @isaaclombardssw @bettybondoc 

Stemming from https://github.com/tinacms/tina.io/pull/2689, when all expandable items in the footer are opened we get sort of a mega menu. This creates a really long footer. 

**Tasks**
- [ ] Evaluate with Betty if we should only allow one accordion open at a time 

![Image](https://github.com/user-attachments/assets/32958774-0ad9-4e0e-b55a-4c7c03205427)

**Figure: Mega Footer when all Accordions open**

bettybondoc:
Hi @isaaclombardssw, @joshbermanssw 

I agree, the scroll is a bit too long and could increase cognitive load.
Let's only allow one accordion open at a time.
Please make sure we save the expanded/collapsed state if users navigate away and return to the page.

joshbermanssw:
Done and live

PR #2745, opened 2025-01-15
💄 Avoid Mega-Menu in Footer

joshbermanssw:
As per #2695, this PR ensures that we do not create a mega menu in our footer by only allowing for one LinkGroup to be open at any one time


isaaclombardssw (approved):
LGTM :)

PR #2784, opened 2025-01-23
⬆️ Dependencies - Updates tinacms and [at]tinacms/cli versions

ncn-ssw:
This updates the tinacms and [at]tinacms/cli packages to confirm their improved stability with a large website.


tino-liu (approved):
Awesome stuff

PR #2815, opened 2025-01-31
🐛Fix Header click behaviour

joshbermanssw:
Fixed the doc and blog scroll behaviour by using an offset and the `window.scrollTo()` function to custom set our offset from the "top of the page"

https://github.com/user-attachments/assets/0a2c14ae-5fc4-4afd-8770-0235fbe150d7
**Figure: New Scroll Behaviour**


isaaclombardssw (approved):
LGTM

PR #2816, opened 2025-01-31
🐛 drop height of left-hand-container to fix scroll height 

joshbermanssw:
As per https://github.com/tinacms/tina.io/issues/2790

I have dropped the height of the `LeftHandSideParentContainer` so that it doesnt hug the bottom of the page when the user is scrolling down the page. See video 

https://github.com/user-attachments/assets/f5760a5c-5846-454e-9b0f-db53aba68672

**Figure: Left hand bar doesnt hug bottom anymore**


isaaclombardssw (approved):
LGTM

bug hunt

Issue #2781, opened 2025-01-22
🐛 The issue of the Navbar exceeding boundaries when accessing Tina.io on some mobile devices.


ZenoWang1999:
### 🐛 Describe the Bug

See https://tina.io/
When accessing tina.io using an iPhone 12 Pro, a blank area appears on the right side.
The blank area is caused by the Navbar at the top exceeding the width of screen.

![Image](https://github.com/user-attachments/assets/2470e8ee-d135-497b-8d3e-bb8524e1670b)
**Figure 1: Blanck area appears on the right side**

### 🧑‍💻 Reproduction Steps

1. Access tina.io on Chrome, right-click and select "Inspect," then click "Toggle device toolbar" and switch to iPhone 12 Pro. Alternatively, access it directly on a mobile device.
2. If the device being used has a relatively small width and height, you will notice a white margin on the right side of the screen.
3. You can use the Responsive mode in Chrome to test at which screen ratios the white margin appears.

### 🎯 Expected Behaviour 

The webpage content should fully cover the entire screen, with no components partially displayed.

### 💼 Tasks

- [ ] Modify the code to ensure the white margin no longer appears.
- [ ] Test across multiple mobile screen ratios to verify that the issue is properly resolved.

### ✅ Acceptance Criteria 

The white margin no longer appears on different mobile screen ratio.

joshbermanssw:
Fixed and live 


![Image](https://github.com/user-attachments/assets/0aba8fd9-5d64-4258-ba1c-6ce78e50789d)

**Figure: White Strip gone**

PR #2820, opened 2025-01-31
🐛Fix iphone white strip

joshbermanssw:
As per https://github.com/tinacms/tina.io/issues/2781 i investigated why the white strip on the iPhone was appearing - it seemed it was from 2 things: 

1. The navbar buttons were too large and causing an x-overflow
2. the RecentPosts section was w-full and then adding padding ontop

**Fixes:**
1. added an extraSmall class for the navbar buttons, this allowed us to drop the size of the buttons without impacting other components on the site 
2. dropped w-full and introduced specific padding for the recentPosts section 

Screenshot 2025-01-31 at 2 36 31 pm

**Figure: navbar fits on mobile now**

Screenshot 2025-01-31 at 2 40 03 pm

**Figure: RecentPosts fits on mobile now**

### DIFF
--- a/components/AppRouterMigrationComponents/AppNavBar.tsx
+++ b/components/AppRouterMigrationComponents/AppNavBar.tsx
@@ -144,7 +144,7 @@ export function AppNavBar({ sticky = true }) {
                 <Button
                   key={index}
                   color={item.color as ValidColors}
-                  size="small"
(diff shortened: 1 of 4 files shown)


TinaInfo (approved):
☑️

PR #2943, opened 2025-03-05
Redesign the route rule

ZenoWang1999:
### General Contributing:

- [ ] Have you followed the guidelines in our [Contributing document](https://github.com/tinacms/tina.io/blob/master/CONTRIBUTING.md)?
- [ ] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?

### All New Content Submissions: (To be confirmed by reviewer)

- [ ] Title is short & specific
- [ ] Headers are logically ordered & consistent
- [ ] Purpose of document is explained in the first paragraph
- [ ] Procedures are tested and work
- [ ] Any technical concepts are explained or linked to
- [ ] Document follows structure from templates
- [ ] All links work
- [ ] The spelling and grammar checker has been run
- [ ] Graphics and images are clear and useful
- [ ] Any prerequisites and next steps are defined.


tino-liu (approved):

middleware.ts
```diff
@@ -7,67 +7,36 @@ export enum SupportedLocales {
...
-  if (!isValidPath) {
-    return;
-  }
-  if (pathname.startsWith('/_next') || pathname.includes('.')) {
-    return;
-  }
-
-  //Show the original info
-  console.log(`default language: ${DEFAULT_LOCALE}`);
+  console.log(`Pass through middleware`);
```
joshbermanssw, line 26: remove log please :)

middleware.ts
```diff
@@ -7,67 +7,36 @@ export enum SupportedLocales {
...
   console.log(`current path: ${pathname}`);
+  console.log(`default language: ${DEFAULT_LOCALE}`);
```
joshbermanssw: rm log

joshbermanssw (changes requested):
https://github.com/user-attachments/assets/a0bd7f38-bebd-4a7f-99ed-35706f042980


I went to chinese - went to the pricing page and it displayed the english pricing page but chinese navigation bar. Then i went to docs and there was no language select. 

1. Fix Chinese pricing page nav 
2. Add lang select to the docs nav bar (not sure why its not there)

ZenoWang1999:
> Screen.Recording.2025-03-11.at.2.57.31.pm.mov 
> I went to chinese - went to the pricing page and it displayed the english pricing page but chinese navigation bar. Then i went to docs and there was no language select.
> 
> 1. Fix Chinese pricing page nav
> 2. Add lang select to the docs nav bar (not sure why its not there)

@joshbermanssw 
1. Because we currently don't have a Chinese pricing page, this needs to be further discussed with colleagues responsible for China marketing. There are two current solutions: 
- Display the English interface directly 
- Show a "not found" page and provide options to jump to the English page
2. This is a design-level issue, as the docs don't support multiple languages at all yet. Therefore, we only provide language-switching buttons on pages that support multiple languages

joshbermanssw:
> > Screen.Recording.2025-03-11.at.2.57.31.pm.mov
> > I went to chinese - went to the pricing page and it displayed the english pricing page but chinese navigation bar. Then i went to docs and there was no language select.
> > 
> > 1. Fix Chinese pricing page nav
> > 2. Add lang select to the docs nav bar (not sure why its not there)
> 
> @joshbermanssw
> 
> 1. Because we currently don't have a Chinese pricing page, this needs to be further discussed with colleagues responsible for China marketing. There are two current solutions:
> 
> * Display the English interface directly
> * Show a "not found" page and provide options to jump to the English page
> 
> 2. This is a design-level issue, as the docs don't support multiple languages at all yet. Therefore, we only provide language-switching buttons on pages that support multiple languages

1. Because we currently don't have a Chinese pricing page, this needs to be further discussed with colleagues responsible for China marketing. There are two current solutions

We shouldnt allow the english interface with the chinese navigation bar i dont believe, so take to a not done page and the option to change back to english 

2. This is a design-level issue, as the docs don't support multiple languages at all yet. Therefore, we only provide language-switching buttons on pages that support multiple languages

I dont think we should show two languages on the same page, might need further opinion but if the docs content is in english so should the rest of the page, unless there is an option to switch back.

ZenoWang1999:
> > > Screen.Recording.2025-03-11.at.2.57.31.pm.mov
> > > I went to chinese - went to the pricing page and it displayed the english pricing page but chinese navigation bar. Then i went to docs and there was no language select.
> > > 
> > > 1. Fix Chinese pricing page nav
> > > 2. Add lang select to the docs nav bar (not sure why its not there)
> > 
> > 
> > @joshbermanssw
> > 
> > 1. Because we currently don't have a Chinese pricing page, this needs to be further discussed with colleagues responsible for China marketing. There are two current solutions:
> > 
> > 
> > * Display the English interface directly
> > * Show a "not found" page and provide options to jump to the English page
> > 
> > 
> > 2. This is a design-level issue, as the docs don't support multiple languages at all yet. Therefore, we only provide language-switching buttons on pages that support multiple languages
> 
> 1. Because we currently don't have a Chinese pricing page, this needs to be further discussed with colleagues responsible for China marketing. There are two current solutions
> 
> We shouldnt allow the english interface with the chinese navigation bar i dont believe, so take to a not done page and the option to change back to english
> 
> 2. This is a design-level issue, as the docs don't support multiple languages at all yet. Therefore, we only provide language-switching buttons on pages that support multiple languages
> 
> I dont think we should show two languages on the same page, might need further opinion but if the docs content is in english so should the rest of the page, unless there is an option to switch back.

@joshbermanssw 
1. Currently, when visiting the pricing page from the Chinese Navbar, users would be directed to /zh/pricing, which shows a Chinese "not found" page. I've now added multilingual support to the not found page.
2. When users visit the docs from the Chinese Navbar, since we don't have a Chinese version available, the English content is displayed and the Navbar will switch to English. Now when clicking on dropdown items in the English Navbar, users are directed to English versions of the content (except the home page, which is determined by cookie preferences). Visiting the docs section doesn't change the user's language preference, as they didn't intentionally choose English content. 

Regarding the language switching button in the docs section, we currently have two options:

1. Hide the language switching button on all interfaces that don't support multiple languages (current code). 
2. Display the button, but for unsupported languages, it would show the not-found page directly instead of trying to search for the English version and returning redirection buttons. The reason is that in Tina, searching for different content types is executed by different query functions, and our not-found page currently only supports the "page" type option.

next.config.js
```diff
@@ -102,6 +110,7 @@ const config = {
     ];
   },
   async redirects() {
+    //TODO We need a better way to handle redirects
```
joshbermanssw, line 113: If we add a TODO can you link a PBI to it  (https://www.ssw.com.au/rules/code-commenting/) otherwise this TODO will get lost. I also dont understand it - the PBI will help me

joshbermanssw (changes requested):
Other than the small reccomendation the only other thing i could see being of value is a toast/ alert when travelling from /zh/* routes to /docs indicating the docs is only in English

joshbermanssw (changes requested):
2 things: 
1. Alert should only be displayed for a few seconds (not be permanent, at the minimum should be clearly closable - 'x' icon) 
2. (not too sure about this one) Alert should only display when coming from Chinese route

joshbermanssw (approved):
as per our conversation, you will be creating improvements for the following in a new PBI 

I spoke to some people in the Sydney office and the tab should remain until it is closed, could you please add an 'x' icon to the tab so that it closes when someone clicks on it (UX) 
Can we only add the tab for when people come from a /zh/** route

ZenoWang1999:
> as per our conversation, you will be creating improvements for the following in a new PBI
> 
> 1. Could you please add an 'x' icon to the tab so that it closes when someone clicks on it (UX)
> 2. Can we only add the tab for when people come from a /zh/** route

See PBI - https://github.com/tinacms/tina.io/issues/2976

PR #2988, opened 2025-03-12
TinaCon - UI changes to Agenda

jeoffreyfischer:
Updated the Agenda style and added a Break in the agenda from 8am to 9am


isaaclombardssw (approved):
✅

PR #3010, opened 2025-03-18
Tutorial beginner guide

isaaclombardssw:
This is a **big PR** which adds tutorials into the Docs. Check out the staging deployment and go to the /docs pages to see the new Learn UI.

Pretty proud of this one, but also it's not exhaustive and should be further developed – content written by Luke C.


app/search/search-client.tsx
```diff
@@ -10,12 +10,15 @@ import {
...
 
 export default function SearchPageClient() {
-  const [props, setProps] = useState<{ formatted?: any }>({});
-
+  const [props, setProps] = useState<{
+    formatted?: any;
+    formattedLearn?: any;
+  }>({});
   const [query, setQuery] = useState('');
+  const [learnActive, setLearnActive] = useState(false);
```
joshbermanssw, line 21: APOC - if we have created the awesome hook for the context why are we using a state again here?

joshbermanssw (approved):
just the one comment - bloody sick job!

app/search/search-client.tsx
isaaclombardssw: using the context 👍

PR #3039, opened 2025-03-20
✨ Tina con banner and button

jeoffreyfischer:
Added banner and registration button to the eventbrite link

We ended up using the PNG graphic instead of creating it as we struggled to anchor the numerous relative items 1:1 to the graphic provided by Pravin. 

Banners have been test passed


app/conference/ConferenceClient.tsx
```diff
@@ -8,23 +8,153 @@ import { GoPeople } from 'react-icons/go';
...
+        <div className="relative hidden lg:flex items-center justify-between px-8 py-6 lg:h-52">
+          <div className="w-[4rem]">
+            <Image
+              src="/svg/logos/tina-white.svg"
+              alt="Tina Logo"
+              width={100}
+              height={100}
+              className="w-[100px] h-auto"
+            />
+          </div>
```
isaaclombardssw, line 45: Should we remove all this stuff (if we're using a png instead?)

isaaclombardssw (changes requested):
Mostly good, but the mobile view is kinda jumpscare with no padding.
SCR-20250324-injy

app/conference/ConferenceClient.tsx
joshbermanssw: As per our conversation - adding a TODO to remove just incase we decide to revert to using this code

isaaclombardssw (approved):
As per my conversation with Josh, good to go

PR #3063, opened 2025-03-24
Update Footer.tsx to have some margin-top

tiagov8:
cc @adamcogan @bettybondoc 

To add breathing space here:

Screenshot 2025-03-24 at 2 45 35 PM

Link: https://tina.io/media


Calinator444:
@tiagov8 

This doesn't seem 

image

tiagov8:
Hey @Calinator444 , I am not running it locally, but this seem like a straight forward change... margin top to the footer div. 
It works fine when using inspect.

Can you check the reason you aren't seeing it isn't a CSS compiling problem?

isaaclombardssw:
Closing as this css change isn't the right approach with our implementation

PR #3067, opened 2025-03-25
update higher def images to tina.io

joshbermanssw:
cc: @BrookJeynes 

Updated conf banner images to higher def


isaaclombardssw (approved):
LGTM

PR #3323, opened 2025-04-24
Moving editorial-workflow doc

sethdaily:
Currently it's in /drafts
It makes more sense for it to be in /tina-cloud.


isaaclombardssw (approved):
LGTM

PR #3339, opened 2025-04-28
Fixing tina conf details

joshbermanssw:
Removing Mish, and adding Calum


isaaclombardssw (approved):
LGTM. So good to have Calum 🔥

PR #3405, opened 2025-05-09
Update navMenu.json

tiagov8:
As per @adamcogan 's request

- Renamed "resources" to "about"
- Moved "about" to second last
- Added apostrophe to "What's" terms


adamcogan (approved):
With Tiago and Marina

PR #3423, opened 2025-05-13
right to down chev

joshbermanssw:
As per https://github.com/tinacms/tina.io/issues/3162 

- [ ] Change Chev Right with transform 90º to Chev down with no transform 

Screenshot 2025-05-14 at 9 17 55 am

**Figure: New Nav Bar**


isaaclombardssw (approved):
✅

Issue #2821, opened 2025-01-31
✨ Replace old llama mascot with Tina.io mascot


isaaclombardssw:
### This has been split between the TinaCloud and Tina.io sides. See https://github.com/tinacms/tinacloud/issues/2566

Requested by: @BenDunkerleySSW via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: @Marxoz, @isaaclombardssw, @adamcogan, @joshbermanssw, @bettybondoc, @wicksipedia

Hi Team!

[Watch the video](https://sswcom.sharepoint.com/:v:/s/SSWYakShaver/EZF9lGkXbTlBtxl5hVG_eRoB3Q3enaolunbFA8KLYad8EQ)

Url: [https://app.tina.io/projects/6c3f3808-fb8f-4bf7-a710-7dfddcda6f61/overview](https://app.tina.io/projects/6c3f3808-fb8f-4bf7-a710-7dfddcda6f61/overview)

### Pain
Replace the old llama mascot on the TinaCMS Contentful Comparison page with the traditional Tina.io mascot. Ensure that the new mascot is consistent with the current theme of Tina.io. Also, check other pages on the Tina.io website for the old llama mascot and replace it where necessary.

### Acceptance Criteria
1. The old llama mascot on the TinaCMS Contentful Comparison page is replaced with the traditional Tina.io mascot.
2. All other instances of the old llama mascot on the Tina.io website are replaced with the traditional Tina.io mascot.
3. The new mascot is consistent with the current theme of Tina.io.

### Screenshot
![Screenshot](https://sayakshaverproduction.blob.core.windows.net/images/Images/01NT34MCN5EUYC7IJ26ZGIWM5J3OJSU2NG_frame_12.png?sv=2024-08-04&se=2075-01-21T13%3A07%3A43Z&sr=b&sp=r&sig=XhYpYjySU%2FwN3q48OSTI80Fr5ou4Pdr66UcK5i48%2FfY%3D)
**Figure: Decision made to replace the old llama mascot with the traditional Tina.io mascot on various pages of the website.**

isaaclombardssw:
Betty:

> FYI - I don't think we have a new _official_ mascot yet. You may want to enlist Ken's help on illustrating this one 😊

Marxoz:
Moving to TinaCloud

wicksipedia:
closing as duplicate

wicksipedia:
> Moving to TinaCloud

[...]

PR #3547, opened 2025-06-10
new mascot for flying

joshbermanssw:
Temporarily putting in a new flying mascot as per https://github.com/tinacms/tina.io/issues/2821


sethdaily (approved):
Llooks good to me

PR #3804, opened 2025-08-04
Adding Security.txt

joshbermanssw:
as per Robs email


isaaclombardssw (approved):
LGTM

public/.well-known/security.txt
```diff
@@ -0,0 +1,7 @@
+Contact: mailto:security@ssw.com.au
```
wicksipedia, line 1: Why not security@tina.io ?

Issue #3963, opened 2025-10-04
✨ Tina.io - Main Menu Restructure


tiagov8:
Requested by: @tiagov8 via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: Daniel Hyles @dotnetdann, @wicksipedia, @tiagov8, @adamcogan, @Marxoz

Hi Team!

&nbsp;&nbsp;[🟥&nbsp;&nbsp;Watch the video (3 min 40 sec)](https://www.youtube.com/watch?v=6Tni9gKOg2M)

&nbsp;&nbsp;Url from screen share: [https://tina.io](https://tina.io)

### Pain
Restructure the main menu on the Tina.io website to improve clarity and user navigation. The current 'Products' menu will be replaced with top-level items for 'TinaCMS', 'Tina Docs', and 'TinaCloud'. Sub-items will be organized under each top-level menu. Pricing and Enterprise will be consolidated, and the 'About' section will be simplified.

### Acceptance Criteria
1. Replace 'Products' menu with top-level items: 'TinaCMS', 'Tina Docs', and 'TinaCloud'.
2. Organize sub-items under each top-level menu: 'Docs', 'What's New', 'Compare', etc.
3. Consolidate 'Pricing' and 'Enterprise' under a single 'Pricing' menu.
4. Simplify 'About' section to 'About Tina'.
~5. Remove 'CMS' from the logo, leaving just 'Tina'.~ replaced by https://github.com/tinacms/tina.io/issues/3971
6. Move 'Videos' under 'TinaCMS'.
7. Collect and analyze user navigation data before and after the menu change.

### Screenshot
![Screenshot](https://sayakshaverproduction.blob.core.windows.net/images/images/ead103a2-6a38-4ceb-a23e-0a790cb398c6_frame_16.png?sv=2025-05-05&se=2045-10-04T02%3A41%3A44Z&sr=b&sp=rd&sig=fcG0gf2Xs4uVpPkUyZCXnVogae24eQPOW5QtZK%2B2YOA%3D)
**Figure: Confusing main menu structure with products listed under a single category.**

tiagov8:
## Menu change

FROM

> - **Products**
>   - [TinaCMS](https://tina.io/tinacms) • [Docs](https://tina.io/docs)
>   - [TinaDocs](https://tina.io/tinadocs) • [Docs](https://tina.io/tinadocs/docs)
> - **[Cloud Pricing](https://tina.io/pricing)**
> - **Community**
[...]

PR #3966, opened 2025-10-06
Update main menu structure for Tina.io

tiagov8:
cc @adamcogan @joshbermanssw @wicksipedia 

Closes https://github.com/tinacms/tina.io/issues/3963

### DIFF
--- a/content/navigationBar/navMenu.json
+++ b/content/navigationBar/navMenu.json
@@ -63,7 +63,7 @@
       "label": "Pricing",
       "items": [
         {
-          "label": "Tina Cloud Pricing",
+          "label": "TinaCloud Pricing",
           "href": "/pricing"
         },
         {
--- a/content/navigationBar/navMenu.json
+++ b/content/navigationBar/navMenu.json
@@ -18,6 +18,10 @@
           "href": "/compare-tina"
         },
         {
+          "label": "Showcase",
+          "href": "/showcase"
+        },
+        {
           "label": "What's New in TinaCMS",
           "href": "/whats-new/tinacms"
         }
@@ -59,10 +63,6 @@
     {
       "label": "Community",
       "items": [
-        {
-          "label": "Showcase",
-          "href": "/showcase"
-        },
         {
           "label": "Roadmap",
           "href": "/roadmap"
--- a/content/navigationBar/navMenu.json
+++ b/content/navigationBar/navMenu.json
@@ -3,8 +3,11 @@
   "navItem": [
     {
       "label": "TinaCMS",
-      "href": "/tinacms",
       "items": [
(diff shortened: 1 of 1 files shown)


tiagov8:
cc @adamcogan @wicksipedia 

Hi @joshbermanssw 

See email: **Tina.io navigation - Remove extra hot fix code**

I recommend we remove the extra code we added to create these "right" and "left" submenu items. 

I think `"_template": "doubleNavItemDropDown"`

They are bad UX and should not be used again (correct if I am wrong). So let's keep the code cleaner.

Screenshot 2025-10-06 at 12 23 50 PM

**❌ Figure: Bad example - Bad UX**

Screenshot 2025-10-06 at 12 24 08 PM

**❌ Figure: Bad example - [Code](https://github.com/tinacms/tina.io/blob/07542bbf34dd5dc4a3ba09178b371fed5fc27107/tina/collectionsSchema/navigationBar.tsx) added to generate the bad UX**

tiagov8:
> 2 things I see issues with - 1. Why are we putting dropdowns for single link items? and 2. How can I now get to /pricing, or /tinadocs, or /tinacms from the nav bar?
> 
> Screenshot 2025-10-07 at 10 13 29 am Screenshot 2025-10-07 at 10 13 26 am
> **Figure: Single Items under dropdown seems wrong**

Thanks for reporting this @joshbermanssw . Fixed in the PR as per our call with @adamcogan and @wicksipedia

brookjeynes-ssw:
I just want to add my 2 cents here and say that it feels confusing to have "TinaCMS" and "TinaDocs". In my opinion, "TinaDocs" shouldn't be in the main header - it should be on the tinadocs website instead.. As a developer, I would click TinaDocs thinking it was the documentation for TinaCMS. TinaDocs is already an action point on the homepage, does it need to be in the header?

+1 for @joshbermanssw comment about drop downs with 1 item in them.

content/navigationBar/navMenu.json
```diff
@@ -2,47 +2,84 @@
...
+          "href": "/whats-new/tinacloud"
+        }
+      ],
+      "_template": "groupOfStringItems"
+    },
+    {
+      "label": "Pricing",
+      "items": [
+        {
+          "label": "Tina Cloud Pricing",
```
joshbermanssw: TinaCloud is one word 
```suggestion
          "label": "TinaCloud Pricing",
```

joshbermanssw (changes requested):

tiagov8:
@brookjeynes-ssw 
> +1 for @joshbermanssw comment about drop downs with 1 item in them.

We have fixed this, thanks for the feedback

PR #4113, opened 2025-11-13
update v of biome

joshbermanssw:
as per title, updating version of biome

### DIFF
--- a/app/blog/page/[page_index]/page.tsx
+++ b/app/blog/page/[page_index]/page.tsx
@@ -45,7 +45,7 @@ export default async function BlogPaginationPage({
   const contentDir = './content/blog/';
   const posts = await glob(`${contentDir}**/*.mdx`);
   const numPages = Math.ceil(posts.length / POSTS_PER_PAGE);
-  const pageIndex = parseInt(params.page_index) || 1;
+  const pageIndex = parseInt(params.page_index, 10) || 1;
   const startIndex = (pageIndex - 1) * POSTS_PER_PAGE;
 
   let postResponse = null;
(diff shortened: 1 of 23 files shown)


app/blog/page/[page_index]/page.tsx
```diff
@@ -45,7 +45,7 @@ export default async function BlogPaginationPage({
   const contentDir = './content/blog/';
   const posts = await glob(`${contentDir}**/*.mdx`);
   const numPages = Math.ceil(posts.length / POSTS_PER_PAGE);
-  const pageIndex = parseInt(params.page_index) || 1;
+  const pageIndex = parseInt(params.page_index, 10) || 1;
```
Comment (Calinator444) [inline app/blog/page/[page_index]/page.tsx:48]: Why do we need to explicitly state the pageIndex is base 10?

It'll default to base 10 unless the index is something like "0x10"

app/zh/blog/page/[page_index]/page.tsx
```diff
@@ -45,7 +45,7 @@ export default async function BlogPaginationPage({
   const contentDir = './content/blog-zh/';
   const posts = await glob(`${contentDir}**/*.mdx`);
   const numPages = Math.ceil(posts.length / POSTS_PER_PAGE);
-  const pageIndex = parseInt(params.page_index) || 1;
+  const pageIndex = parseInt(params.page_index, 10) || 1;
```
Comment (Calinator444) [inline app/zh/blog/page/[page_index]/page.tsx:48]: Again, not sure why we're passing the radix parameter here

https://github.com/tinacms/tina.io/pull/4113/files#r2525047362

app/blog/page/[page_index]/page.tsx
Comment (joshbermanssw) [reply app/blog/page/[page_index]/page.tsx:48]: As per our convo - linting

app/zh/blog/page/[page_index]/page.tsx
Comment (joshbermanssw) [reply app/zh/blog/page/[page_index]/page.tsx:48]: See above

Issue #4057, opened 2025-10-31
✨ Footer & Community page - Add comments box to subscription forms


GriffenEdge:
Requested by: @griffenedge via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: @griffenedge, @micsblank, @adamcogan, @Marxoz, @wicksipedia, @joshbermanssw 

Hi Team!

&nbsp;&nbsp;[🟥&nbsp;&nbsp;Watch the video (39 sec)](https://www.youtube.com/watch?v=8nvkn96BTuM)

&nbsp;&nbsp;Url from screen share: [https://tina.io/community](https://tina.io/community)

### Pain
Add a comments box to the end of the forms on the Tina.io community page, following the fields for first name, last name, and email.

### Acceptance Criteria
1. A comments box is added to the end of the specified forms on the Tina.io community page.
2. The comments box appears after the fields for first name, last name, and email.
3. The comments box is a text area form.

### Screenshot
![Screenshot](https://sayakshaverproduction.blob.core.windows.net/images/images/ea194d34-0845-4004-9559-eeb3f7df7838_frame_15.png?sv=2025-05-05&se=2045-10-31T04%3A14%3A16Z&sr=b&sp=rd&sig=7Kxj%2BEkfM7K%2B0T6fprJ7%2F3aD9NxxC7CaO10Hjs%2F8C7w%3D)
**Figure: Add comments box to forms**

griffenedge:
**<figure><img width="2880" height="1920" alt="" src="https://github.com/user-attachments/assets/9b78b62f-9630-4c8a-b34d-ffe6d00ae727" /><figcaption>Figure: Mockup of new notes field</figcaption></figure>**

joshbermanssw:
✅ Done - 

<img width="326" height="367" alt="Image" src="https://github.com/user-attachments/assets/1196e8ad-b24c-4053-8f5d-85f7f767f3ef" />

PR #4117, opened 2025-11-14
add notes field to subscription form 

joshbermanssw:
cc: @GriffenEdge 

As per https://github.com/tinacms/tina.io/issues/4057

Adding notes field to all of the subscription forms 

Screenshot 2025-11-14 at 5 28 12 pm

**Figure: Notes Field**

Screenshot 2025-11-14 at 5 18 10 pm

**Figure: Data showing in Mailchimp**

### DIFF
--- a/components/modals/AppRouterEmailForm.tsx
+++ b/components/modals/AppRouterEmailForm.tsx
@@ -1,7 +1,7 @@
 'use client';
 import Image from 'next/image';
 import type React from 'react';
-import { useState } from 'react';
+import { useRef, useState } from 'react';
 import { ImCross } from 'react-icons/im';
 import { IoIosWarning } from 'react-icons/io';
 import { TiTick } from 'react-icons/ti';


components/modals/AppRouterEmailForm.tsx
```diff
@@ -1,23 +1,32 @@
...
-  const [email, setEmail] = useState('');
-  const [firstName, setFirstName] = useState('');
-  const [lastName, setLastName] = useState('');
-  const [_isEntering, setIsEntering] = useState(false);
+  const [formData, setFormData] = useState<FormData>({
+    firstName: '',
+    lastName: '',
+    email: '',
+    notes: '',
+  });
```
Calinator444: Hm... 🤔 It might be a good idea to keep the default values of the form in a variable. Just in case they change.

```suggestion
  const defaultValues = useRef<FormData>({
    firstName: '',
    lastName: '',
    email: '',
    notes: '',
  });
  const [formData, setFormData] = useState<FormData>(defaultValues.current);
```

components/modals/AppRouterEmailForm.tsx
```diff
@@ -26,12 +35,23 @@ export const EmailForm = (props: EmailFormProps) => {
...
         setMessage({
           text: "You've been added to the llama list.",
           type: 'success',
         });
+        setFormData({
+          firstName: '',
+          lastName: '',
+          email: '',
+          notes: '',
+        });
```
Calinator444: Now we can re-use the default values
```suggestion
        setFormData(defaultValues.current);
```

Calinator444:
Looking good :)

PR #4122, opened 2025-11-17
change faq from docs/intro/faq to docs/faq and add redirect

joshbermanssw:
cc: @adamcogan @wicksipedia 

Hey Josh
1. Investigate with Wicksy and see if we should move /docs/introduction/faq
    to the simpler
	/docs/faq
2. Can you put a redirect
3. Are you happy with the order?


amankumarrr (changes requested):
can we add this through UI? 
image

amankumarrr (approved):

PR #4164, opened 2025-12-01
Delete AppRouterMigrationComponents Directory

joshbermanssw:
As per https://github.com/tinacms/tina.io/issues/3934

There was a temporary readme (see at bottom) with instructions to remove a temporary directory after the migration to the app router was complete. This was completed a while ago 😅 

**Changes**
- Deleted unused components in `components` directory
- Move all used components from the AppRouterMigrationComponents directory to the components directory 

-----

**TEMPORARY README**

The directory `components/AppRouterMigrationComponents/` contains components currently used in the
pages router implementation of the site but modified to fit the client server pattern of the app router.

All components created, and used in the AppRouter implementation are temporarily stored in this directory

Once the migration from pages to app router is finished the old components in the `components/` directory will be deleted and the components from `components/AppRouterMigrationComponents/` will be moved out of the `AppRouterMigrationComponents/` directory back in to the `components/` directory so that there are **no duplicate components**


public/robots.txt
Calinator444, line 1: Why is the sitemap updated?

content/docs/index.mdx
```diff
@@ -7,9 +7,10 @@ seo:
     TinaCMS is an open-source, Git-backed CMS with real-time editing, version
     control, and scalability, empowering developers and content creators alike.
 title: What is TinaCMS?
-last_edited: '2025-06-19T06:47:19.122Z'
+last_edited: '2025-12-01T01:34:24.084Z'
```
Calinator444, line 10: This looks like a content change. Is this necessary?

components/AppNavBar.tsx
```diff
@@ -20,8 +20,8 @@ import { Modal } from 'react-responsive-modal';
 import { getGitHubStarCount } from '@/utils/github-star-helper';
 import { saveLocaleToCookie } from '@/utils/locale';
 import 'react-responsive-modal/styles.css';
-import { EmailForm } from '../modals/AppRouterEmailForm';
-import { Button } from '../ui/Button';
+import { EmailForm } from './modals/AppRouterEmailForm';
```
Calinator444: Can you rename this "AppRouterEmailForm" please? From PR description it looks like there should only be 1 variation of this file

app/zh/blog/[...slug]/BlogPageClient.tsx
```diff
@@ -1,7 +1,7 @@
 'use client';
 
 import Giscus from '@giscus/react';
-import { formatDate } from 'components/AppRouterMigrationComponents/utils/formatDate';
+import { formatDate } from '@/component/util/formatDate';
```
Comment (Calinator444) [inline app/zh/blog/[...slug]/BlogPageClient.tsx]: Should this be in `components`? judging from the name it looks like it's a utility method

content/docs/index.mdx
joshbermanssw: No - when the page is loaded in the admin tab it automatically updates last_edited

app/zh/blog/[...slug]/BlogPageClient.tsx
Comment (joshbermanssw) [reply app/zh/blog/[...slug]/BlogPageClient.tsx]: moved to utils dir

content/docs/index.mdx
Calinator444: WHYYYY 

Nah but that's okay

public/robots.txt
Calinator444: It looks like this was deleted, so all good.

Calinator444:
Looks good to me

Issue #4179, opened 2025-12-03
✨ Enhanced Link Interaction - Improve User Guidance After Copying


joshbermanssw:
Requested by: @joshbermanssw via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: @SamBWagner, @18-th, @joshbermanssw, @Marxoz, @wicksipedia, @adamcogan

Hi Team!

&nbsp;&nbsp;[🟥&nbsp;&nbsp;Watch the video (1 min 53 sec)](https://www.youtube.com/watch?v=AptMq5U0j4U)

&nbsp;&nbsp;Url from screen share: [https://tina.io](https://tina.io)

### Pain
Enhance the user experience by providing additional guidance and links after the 'copy to clipboard' action is performed. This includes linking to relevant documentation and videos for troubleshooting and setup instructions.

### Acceptance Criteria
1. After clicking 'copy to clipboard', display a message with links to relevant documentation and videos.
2. Include instructions for opening the terminal app in admin mode and pasting the copied content.
3. Provide troubleshooting steps and common issues related to npm installation and Windows execution policies.
4. Ensure the links direct users to the correct documentation pages with Josh's video and troubleshooting information.

### Screenshot
![Screenshot](https://sayakshaverproduction.blob.core.windows.net/images/images/a28da837-97ba-4bb5-9efe-a632c4def420_frame_20.png?sv=2025-05-05&se=2045-12-03T07%3A58%3A54Z&sr=b&sp=rd&sig=6GgeO6M92YaiD5sHdkkNgNntSayfAXQJM7JdTK34sOI%3D)
**Figure: User requests additional information after clicking 'copy to clipboard'.**

joshbermanssw:
✅ Done https://github.com/user-attachments/assets/bb82e7bc-b41a-401b-b45f-41f1d31214cc

**Figure: Updated UI**

PR #4183, opened 2025-12-05
add need help popover

joshbermanssw:
As per #4179 

Added an optional rich-text field to the code copy button that allows users to add text to a pop-over field. This pop-over field uses a sessionKey to track whether it should be open or not. This helps guides users down "no wrong path" if they click on it, need help, but then refresh the page

https://github.com/user-attachments/assets/bb82e7bc-b41a-401b-b45f-41f1d31214cc

**Figure: Updated UI**

### DIFF
--- a/content/docs-toc/docs-toc.json
+++ b/content/docs-toc/docs-toc.json
@@ -29,10 +29,10 @@
           "_template": "item"
         },
         {
-          "title": "Supported Frameworks",
+          "title": "Getting Started with Tina",
           "items": [
             {
-              "title": "Starter Kits",
+              "title": "Tina Starter - CLI",
               "slug": "content/docs/introduction/using-starter.mdx",
               "_template": "item"
             },
--- a/content/docs/introduction/using-starter.mdx
+++ b/content/docs/introduction/using-starter.mdx
@@ -1,7 +1,7 @@
 ---
 id: /docs/introduction/using-starter/
-title: Starter Kits
-last_edited: '2025-11-11T06:14:54.467Z'
+title: Tina Starter - CLI
+last_edited: '2025-12-07T21:20:13.520Z'
 next: content/docs/frameworks/next/app-router.mdx
 previous: content/docs/vibe-coding.mdx
 ---


SamBWagner:
I think this is a marked improvement. But, I think it could be better.

Let's simplify it to just be some grey text that appears when you click it.

Reasoning is that it's appearing as a Pop-up, which reads as something that is not persistent. So, now the session storage solution where it's persistent reads as a bug.

So, change

the Pop-up

to

Some simple grey text/a link.

joshbermanssw:
Updated UI 

https://github.com/user-attachments/assets/db68045c-c0b9-41a4-bd5d-d6b2ba77cc06

components/blocks/CodeButton/CodeButton.template.ts
```diff
@@ -10,5 +10,14 @@ export const codeButtonTemplate = {
...
-  fields: [{ name: 'label', label: 'Label', type: 'string' }],
+  fields: [
+    { name: 'label', label: 'Label', type: 'string' },
+    {
+      name: 'popOverText',
+      label: 'Pop Over Text',
+      type: 'rich-text',
+      description:
+        'Optional:This text will appear when the user clicks on the button',
+    },
```
SamBWagner, line 21: I'm not someone who understands Tina's custom component templating but I think the name "popOverText" should be changed as it's no longer a dialog/popup. Rather it's more like "inlineHelp" or "assistiveText" or "helpText". I'd change it to one of those three that you think fits best.

components/blocks/CodeButton/CodeButton.tsx
```diff
@@ -24,46 +25,70 @@ export const CodeButton = ({
   label,
   id,
   className = '',
+  popOverText,
```
SamBWagner: Same as the above suggestion, just in the component itself.

content/blocksPages/home.json
```diff
@@ -30,9 +30,11 @@
         },
         {
           "label": "npx create-tina-app@latest",
+          "popOverText": "Need some help? Check out the [docs](/docs/introduction/using-starter)\n",
```
SamBWagner: Samesies

content/blocksPages/tinacms.json
```diff
@@ -28,9 +28,11 @@
         },
         {
           "label": "npx create-tina-app@latest",
+          "popOverText": "",
```
SamBWagner: Samesies

utils/renderButtonArrayHelper.tsx
```diff
@@ -14,7 +14,12 @@ const RenderButton = ({ button, className = '' }) => {
   } else if (button.__typename.includes('CodeButton')) {
     return (
       <div>
-        <CodeButton id={button.id} label={button.label} className={className}>
+        <CodeButton
+          id={button.id}
+          label={button.label}
+          className={className}
+          popOverText={button?.popOverText}
```
SamBWagner: Samesies

components/blocks/CodeButton/CodeButton.template.ts
joshbermanssw: Resolved

content/blocksPages/home.json
joshbermanssw: Resolved

content/blocksPages/tinacms.json
joshbermanssw: Resolved

utils/renderButtonArrayHelper.tsx
joshbermanssw: Resolved

SamBWagner (approved):
Lgtm

PR #4306, opened 2026-02-05
TinaCMS Lingo 

tiagov8:
Adding more terms from email: **Re: Getting the TinaCMS "Lingo" Right (for future videos and documentation)**

Closes Issue: https://github.com/tinacms/tina.io/issues/4320

as per @joshbermanssw 
Also adding "tinacms-lingo" to the sidebar menu


content/docs/tinacms-lingo.mdx
```diff
@@ -9,16 +10,85 @@ previous: content/docs/setup-overview.mdx
 
 As TinaCMS continues to evolve, it's important to use clear and consistent language when describing its editing experience. We often reference a set of terms that explain how visual editing works and how editors interact with content. The following definitions outline these terms and show how they fit together:
 
-**Visual Editing**\
-The overall two-pane editing experience, with the **Editor** on the left and a live **Site Preview** on the right.
+### Admin Sidebar
+
+The Sidebar with collections listed (Not to be confused with the TinaCMS Editor)
+
+![](/lingo/TinaCMs-Admin-Sidebar.jpeg "Left-hand side Admin Sidebar")
```
wicksipedia: Images should be kebab cased as inconsistent casing can introduce issues. Git is case sensitive with filenames (as Linux is)

content/docs/tinacms-lingo.mdx
```diff
@@ -9,16 +10,85 @@ previous: content/docs/setup-overview.mdx
...
+![](/lingo/Visual-Editing.jpeg "Visual editing with Editor (left) and Site Preview (right)")
+
+### Site Preview
+
 The right-hand pane that displays a live view of the website as content is edited.
 
-**Visual Element Highlighting**\
+### Visual Element Highlighting
+
 Hovering over elements in the **Site Preview** highlights them in blue 🟦, helping identify page structure and content relationships.
```
wicksipedia: The current text mentions "hovering," but in TinaCMS, these blue outlines are constant indicators of editable regions. Additionally, the primary function of these outlines is to allow users to click into the preview to jump directly to the relevant field in the CMS sidebar.

I also think we shouldn't use the term "Highlighting" since these outlines are persistent and don't require a hover state to appear. "Highlighting" implies a temporary visual change that isn't actually happening.

```suggestion
### Editable Regions

Elements in the **Site Preview** with a blue outline 🟦 are editable. Clicking any outlined element will automatically focus its corresponding field in the sidebar for immediate editing.
```

content/docs/tinacms-lingo.mdx
tiagov8: Fixed all images file names and paths

public/lingo/tinacms-admin-sidebar.jpeg
wicksipedia, line 1: Are the large beige borders around the images intentional?

It looks quite strange when seeing it in context on the site

Image

tiagov8: I agree it looks strange. Not sure why they were included in screenshots. Waiting to see what @joshbermanssw had in mind

joshbermanssw: Just capturing like this because its easy to do so with Arc browser. Would you like me to remove? @wicksipedia

wicksipedia: Ok that makes sense why they are there.

IMHO the border is way too fat and the colour looks out of place. It does need some padding but maybe if it was a 5px white border instead it would look better?

tiagov8: Agree with @wicksipedia they look fat and have inconsistent color.

However, I think we should avoid any padding so they are consistent with other images in other Docs pages (e.g. https://tina.io/docs/using-tina-editor)

BTW - Whenever we do want to have paddings/borders, it is preferred to use CSS rather than including in the images themselves

tiagov8: cc @joshbermanssw 

Hi @wicksipedia 

Done - Borders removed from images

wicksipedia (approved):
the updated images aren't showing on the preview, but that should be resolved once merged to `main`

PR #4307, opened 2026-02-05
fix: 'Github' to 'GitHub'

joshbermanssw:
As per #3952 

Fixes spelling of Github to GitHub


tina/tina-lock.json
Calinator444, line 1: This lock file needs to be merged against main. Otherwise LGTM.

Just make sure you re-run `pnpm i` and `pnpm dev` after merging against main. We're running the latest tina version now which changes the lock file.

PR #4477, opened 2026-04-22
Add /versions page listing TinaCMS npm package versions

joshbermanssw:
## Summary
- New `/versions` page, rendered by a TinaCMS **Versions** block. Title, description, and a repeatable **Buttons** array (using the shared Actions template — seafoam, orange, ghost, etc.) are all editable in `/admin`.
- The block fetches each TinaCMS package from `registry.npmjs.org` on the client and renders a responsive table (desktop) / card list (mobile). Row resolves progressively as each fetch returns.
- Shimmer skeletons matched to the exact resolved-row layout so there's no layout shift on first paint. Shared primitive at `components/ui/SkeletonBar.tsx`.
- All coding standards honored: no `<style>` tags, no inline styles, no arbitrary bracket magic numbers — shimmer / row-in animations and the skeleton gradient live in `tailwind.config.js`.
- `fetchPackageInfo` + `formatPublishedDate` (with tests) live in `utils/` — both are generic utilities and don't belong under a specific block.

## Implements
tinacms/tinacms#6715

Screenshot 2026-04-22 at 3 44 56 pm

**Figure: UI**


isaaclombardssw (approved):
Love it. Code looks good. ✅

PR #4581, opened 2026-05-29
Fix overpromising slug on the astro-default-starter blog post

kulesy:
## Why

The blog post's H1 and SEO description read **"Astro is _becoming_ the default starter for TinaCMS"** (future-progressive, intentional — the CLI default isn't flipped yet, the new React-free path is still polishing), but the URL slug reads **`astro-is-the-default-tinacms-starter`** (present tense).

This matters right now because:
- The post has just been linked to the Astro team in [withastro/docs#13936](https://github.com/withastro/docs/pull/13936) as the public reference for their May 31 newsletter
- If they quote the slug, it implies the switch has already happened — which contradicts the body's "we have not changed the default yet" line and the Astro PR thread where we've explicitly said the flip is still weeks away

## What

- Rename `content/blog/astro-is-the-default-tinacms-starter.mdx` → `content/blog/astro-is-becoming-the-default-tinacms-starter.mdx`
- Update `canonicalUrl` in the post's frontmatter
- Update `next:` pointer in the adjacent post `customblog_tinacmsai.mdx`
- Add a 301 redirect in `content/settings/config.json` so the already-shared link doesn't 404

Image directory `/img/blog/astro-is-the-default-tinacms-starter/` left as-is (asset path, not user-facing URL).

## Slug choice

`astro-is-becoming-the-default-tinacms-starter` — exact match to the H1 and SEO title/description. Happy to swap to a different slug if anyone prefers shorter — just suggest one and I'll re-do.

## Test plan

- [ ] Netlify preview renders the new URL correctly
- [ ] Visiting the old URL on the preview 301s to the new one
- [ ] `customblog_tinacmsai.mdx`'s "next post" link points at the new path


18-th (approved):
LGTM

Issue #4592, opened 2026-06-01
Improve tina-io performance: cut build churn, fix image delivery, edge-cache hot routes


joshbermanssw:
## Problem

`tina-io` does a lot of **uncached per-request work** and **redundant rebuilds** that hurt response times, image delivery, and the path-to-production feedback loop. Backed by the Vercel Usage page for the current cycle, `tina-io` is the **top project** for Build CPU and Image Optimization.

## Findings (current billing cycle, from Usage + runtime logs)

- **Build CPU — `tina-io` = 74% of the team total (~52h in 6 days).** Driven by deploy churn: bot translation PRs (`translate-pr-*`), the release-notes bot (`releaseNotes/*`), TinaCMS Cloud content commits, PR pushes — plus repeatedly-**failing** builds that still burn CPU. One production build ≈ 7.5 min.
- **Image Optimization — `tina-io` = 93% of Cache Writes (~501K write units).** Cache Writes > Transformations means images are being **re-optimized/re-written repeatedly** (cache churn / short TTL), not transformed once. Observability shows single images (e.g. `Relax-Llama.png`) hit ~175 transformations in 12h.
- **Function Duration on `tina-io` is comparatively small (~16% of team total).** The hot paths are still worth fixing for TTFB: `/api/githubstar` (nav widget, **no `Cache-Control`**, calls GitHub every request) and `/api/github-metadata` (caches upstream but still runs a function per request). `/_next/image` also shows in serverless logs.
- Constant `HEAD /` (uptime monitor) + full-site `GET` crawler sweeps multiply all of the above.

### 1. Cut Build CPU / speed up deploys (biggest lever — 74% of build is this project)
- [ ] Add a Vercel **Ignored Build Step** so bot branches (`translate-pr-*`, `releaseNotes/*`) and content-only commits don't trigger full ~7.5 min builds.
- [ ] Fix the repeatedly-erroring `tech-debt/docs-blog-i18n-dedup` builds (failed builds still consume build CPU).
- [ ] Confirm superseded preview deployments auto-cancel.
- [ ] Consider batching the auto-generated translation merges instead of one full prod build each.

### 2. Fix image delivery (93% of image cache writes is this project)
- [ ] In `next.config.js` `images`: set a long `minimumCacheTTL` (e.g. 31 days) to stop the cache-write churn.
- [ ] Trim `deviceSizes` / `imageSizes` / `qualities` to what's actually used (fewer variants per image → faster cache hits, better LCP).
[...]

PR #4594, opened 2026-06-01
⚡ Edge-cache GitHub API routes, fix image cache churn, skip bot-branch preview builds

joshbermanssw:
## TL;DR

`tina-io`'s biggest Vercel runtime/build hotspots are uncached per-request work and redundant builds. This caches the two GitHub-backed API routes at the edge, sets `minimumCacheTTL` so optimized images stop being re-written every 60s, and adds an Ignored Build Step (wired via `vercel.json`) that skips the ~7.5 min build for bot-branch *preview* deployments. Addresses the code-level items in #4592.

**Files to review (5, +73 / -8):**

| File | Why |
|---|---|
| `app/api/githubstar/route.ts` | Was fully uncached (function + GitHub call on every page view) — adds `Cache-Control`. |
| `app/api/github-metadata/route.ts` | Cached upstream but ran a function per request — adds matching response cache. |
| `next.config.js` | `minimumCacheTTL` + drop 3840px `deviceSize`. |
| `scripts/vercel-ignore-build.sh` *(new, start here)* | Skips preview builds for `translate-pr-*` / `releaseNotes/*`. |
| `vercel.json` *(new)* | Wires the script in via `ignoreCommand` (version-controlled, overrides the dashboard field). |

## Why

The nav star widget calls `/api/githubstar` on essentially every page render, and it had no cache headers — so every visitor and every crawler hit re-invoked the function and called GitHub. Image optimization's dominant cost was **cache writes, not transforms**: Next's default `minimumCacheTTL` is 60s, so optimized variants were re-fetched and re-written constantly. And bot-generated branches (Chinese translations, release notes) were each triggering full preview builds nobody reviews.

## How

- **API routes:** wrap the success responses with `Cache-Control: public, s-maxage=…, stale-while-revalidate=86400`, so repeat requests are served from the Vercel edge (`x-vercel-cache: HIT`) instead of re-running the function. `github-metadata` reuses its existing `CACHE_TTL` (300s); `githubstar` uses 1h (the count is identical site-wide and slow-moving).
- **Images:** `minimumCacheTTL: 2678400` (31d) forces long reuse of optimized variants; dropping the 3840px variant halves transform/cache cost per image (Next falls back to 2048px).
- **Build skip:** `vercel.json` `ignoreCommand` runs the script, which exits 0 (skip) only for preview deployments on bot branches; production and human PRs always build. Smoke-tested for all four cases.

## Reviewer notes

- **Production always builds.** The script gates on `VERCEL_ENV=preview` + bot ref only; anything else proceeds.
- **Content commits still build.** tina-io builds content into the site, so the skip deliberately targets *bot preview* builds, not content changes.
- **No `qualities` config.** That's Next 15-only; this repo is on 14.2.

## Follow-up (dashboard — can't be done in-repo)

This PR makes the selective-skip work, but two settings still need flipping in Project Settings → Git for the intended preview behavior:

1. **Re-enable non-production auto deployments** (so previews run and the ignore step can selectively skip them).
2. **Enable auto-cancel of superseded deployments** (collapses rapid re-pushes / TinaCMS saves to the latest commit only).

Also out of scope: Firewall bot management, and evaluating a custom image loader for `assets.tina.io`.

## Links

- Closes part of tinacms/tina.io#4592

### DIFF
--- a/app/api/github-metadata/route.ts
+++ b/app/api/github-metadata/route.ts
@@ -75,13 +75,23 @@ export async function GET(request: Request) {
     const latestCommit = latestCommits[0];
     const firstCommit = allCommits.length ? allCommits.at(-1)! : null;
 
-    return NextResponse.json({
-      latestCommit,
-      firstCommit,
-      historyUrl: path
-        ? `https://github.com/${owner}/${repo}/commits/main/${path}`
-        : `https://github.com/${owner}/${repo}/commits/main`,
-    });
+    return NextResponse.json(
+      {
+        latestCommit,
+        firstCommit,
+        historyUrl: path
+          ? `https://github.com/${owner}/${repo}/commits/main/${path}`
+          : `https://github.com/${owner}/${repo}/commits/main`,
+      },
+      {
+        // Cache per (owner, repo, path) at the Vercel edge so repeat requests
+        // are served from cache instead of re-invoking the function. Mirrors
+        // the CACHE_TTL used for the upstream GitHub fetch above.
+        headers: {
+          'Cache-Control': `public, s-maxage=${CACHE_TTL}, stale-while-revalidate=86400`,
+        },
+      },
+    );
   } catch (error) {
     console.error('Error fetching GitHub metadata:', error);
     return NextResponse.json(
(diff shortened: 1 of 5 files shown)


isaaclombardssw (approved):
Looks reasonable

Marxoz (approved):

PR #4613, opened 2026-06-05
docs: FAQ entry for npm Vite version-skew build failure

wicksipedia:
## What

Adds FAQ entry #22 to `content/docs/faq.mdx` covering the `Missing field tsconfigPaths` build failure npm users hit.

## Why

Under **npm**, two major versions of Vite can end up in the dependency tree (e.g. Vite 8 pulled in by `@tailwindcss/vite` alongside the Vite 7 Astro uses), and the mismatched rolldown bindings crash the build. Reported in [tina-astro-starter#58](https://github.com/tinacms/tina-astro-starter/issues/58).

The new entry recommends **pnpm** (which dedupes to a single Vite) and gives the npm `overrides` workaround. Pairs with the create-tina-app change that recommends pnpm by default.

### DIFF
--- a/content/docs/faq.mdx
+++ b/content/docs/faq.mdx
@@ -255,7 +255,7 @@ As a workaround, you can rename media files by making the changes directly in Gi
 
 For more details, see [Renaming Media Files](/docs/reference/media/repo-based/#renaming-media-files).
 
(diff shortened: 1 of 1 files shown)


18-th (approved):
LGTM

PR #4628, opened 2026-06-12
🐛 Fix release notes parser dropping single-space markdown bullets

joshbermanssw:
## TL;DR

The auto-generated release notes page for TinaCMS v3.9.1 (#4612) was published empty — the parser silently dropped every bullet and fell back to `"changesDescription": ""`. The bullet matcher in `create-release-notes.sh` now accepts any whitespace after the list marker, and the script fails loudly instead of writing an empty page.

**Files to review (3, +223 / -19):**

| File | Why |
|---|---|
| `scripts/create-release-notes.sh` *(start here)* | Relaxed bullet matching in both awk blocks + new empty-output guard. |
| `scripts/create-release-notes.test.ts` *(new)* | Jest tests that shell out to the script, including the verbatim v3.9.1 input that broke. |
| `content/whats-new-tinacms/v3.9.1.json` | Regenerated with the fixed script from the original workflow input — restores the missing page content. |

## Why

The awk parser only accepted bullets shaped exactly like `-␣␣␣item` (dash + three spaces, changesets' prettier-formatted style) or `*␣item`. The v3.9.1 release notes arrived with standard single-space bullets (`-␣item`), so the only item hit the `else next` branch and was discarded; with zero items parsed, the script printed its hardcoded empty fallback, exited 0, and the workflow happily opened (and auto-merged) a PR with a blank page. Compare [run 26860023803](https://github.com/tinacms/tina.io/actions/runs/26860023803) (v3.9.0, three-space bullets, parsed fine) with [run 27000040038](https://github.com/tinacms/tina.io/actions/runs/27000040038) (v3.9.1, single-space bullets, empty output).

## How

- Both awk blocks now match `^[-*][[:space:]]` and strip the marker with `gsub` instead of fixed-offset `substr`, so any bullet indentation parses.
- The same relaxation applies to the `Updated dependencies` detection and its nested dep lines, which had the identical rigid pattern.
- A post-write guard counts items and empty descriptions with `jq`; if nothing parsed, the script deletes the file, echoes the raw input, and exits 1 — so the workflow fails visibly instead of auto-merging an empty page.

## Reviewer notes

- **Guard covers both fallback shapes.** Empty descriptions appear as plain `""` or as rich-text objects with empty `text` nodes depending on the code path; the jq check normalizes both.
- **v3.9.1.json regenerated, not hand-written.** Output of the fixed script run with the exact `RELEASE_NOTES` input captured from the original workflow logs.

## Tests

5 new Jest tests in `scripts/create-release-notes.test.ts` run the script end-to-end in a temp dir: single-space bullets (the v3.9.1 regression), three-space bullets (v3.9.0 format), single-space `Updated dependencies` with nested deps, TinaCloud asterisk bullets, and the fail-loudly guard. Run with `npx jest scripts/create-release-notes.test.ts`; full suite (45 tests) passes.

## Links

- [PR #4612 — the empty release notes page this fixes](https://github.com/tinacms/tina.io/pull/4612)
- [Workflow run that produced the empty page](https://github.com/tinacms/tina.io/actions/runs/27000040038)

### DIFF
--- a/components/footer/footer.tsx
+++ b/components/footer/footer.tsx
@@ -147,7 +147,7 @@ export function Footer({ footerData }: { footerData: FooterData }) {
               <TinaIcon color="white" />
             </div>
             <div className="flex-1 grid grid-cols-2 py-2 lg:py-0 md:grid-cols-2 lg:grid-cols-4 gap-6 md:gap-4">
-              {currentFooterNav.map((item, columnIndex) => {
+              {currentFooterNav.map((item, _columnIndex) => {
                 const { header, footerItem } = item;
                 return (
                   <div
--- a/content/whats-new-tinacms/v3.9.1.json
+++ b/content/whats-new-tinacms/v3.9.1.json
@@ -3,10 +3,16 @@
   "dateReleased": "2026-06-05T06:46:13Z",
   "changesObject": [
     {
-      "changesTitle": "Release Notes",
+      "changesTitle": "Patch Changes",
       "changesList": [
         {
-          "changesDescription": ""
+          "changesDescription": "Skip the filesystem-backed response cache on edge runtimes (Cloudflare Workers, Vercel Edge) where Node's `fs` API is present but unusable, which could otherwise hang concurrent identical queries. Adds a `cache` option to `createClient` to force-disable the cache.",
+          "pull_request_number": "7028",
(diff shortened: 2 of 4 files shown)


isaaclombardssw (approved):
fix the build
✅ sure

PR #4632, opened 2026-06-12
🩹 Restore whats-new-tinacms/v3.9.1 to fix the main build

joshbermanssw:
## TL;DR

Production builds on `main` fail prerendering `/whats-new/tinacms` with `Error querying file content/whats-new-tinacms/v3.9.1.json from collection WhatsNewTinaCMS`. This PR reverts #4631 and restores the file so the build passes.

## Why

The TinaCMS admin showed two v3.9.1 documents, but git history shows only one `v3.9.1.json` ever existed — the second was a stale TinaCloud index entry. Deleting "the duplicate" via the admin (#4631) therefore removed the only real file, leaving an orphaned index entry on the `main` branch index. Every `WhatsNewTinaCMSConnection` query now tries to resolve that entry against a file that no longer exists in git, which kills static generation of `/whats-new/tinacms` and the whole deploy.

Restoring the file gives the index entry something to resolve to. The v3.9.1 release notes content is valid and belongs on the site anyway (real release, added in #4612).

## Follow-up

The stale duplicate entry still needs a branch re-index of `main` in the TinaCloud dashboard (app.tina.io) — until then the admin may keep showing two v3.9.1 documents. Do **not** delete the file again; re-index instead.

## Links

- Deletion this reverts: #4631
- Original release notes: #4612

### DIFF
--- a/content/whats-new-tinacms/v3.9.1.json
+++ b/content/whats-new-tinacms/v3.9.1.json
@@ -1,0 +1,20 @@
+{
+  "versionNumber": "3.9.1",
+  "dateReleased": "2026-06-05T06:46:13Z",
+  "changesObject": [
+    {
+      "changesTitle": "Patch Changes",
+      "changesList": [
+        {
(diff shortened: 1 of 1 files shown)


18-th (approved):
LGTM

PR #4645, opened 2026-06-17
Add agency partners page

isaaclombardssw:
No linked issue.

**TL;DR** Adds an agency partners page at `/partners`, with two new blocks and a partner-application variant of the contact form.

**Pain:** We have nowhere to send people who want a TinaCMS build done for them, and nowhere for agencies who want to do that work to put their hand up. Every comparable CMS (Sanity, Strapi, Contentful, Storyblok) has a partners page serving those two audiences; we don't, so both conversations happen ad hoc over email.

**Solution:** A new `content/main/partners.json` page (auto-routed, linked from the nav under Community) built from existing blocks plus two new ones: `partnerGrid`, a card grid of vetted partners that always closes with a "not sure who fits?" contact card, and `imageBand`, a full-width captioned image. Both are registered in `pages.tsx` and `Blocks.tsx`, so the whole page is editable in `/admin`. The apply CTA opens `ContactForm` under a new `partner` variant that collects what we actually need to triage an application (sole developer vs agency, website, team size or availability, TinaCMS experience) and tags the resulting email. Two fixes to that form apply to the navbar and footer instances too: the Radix `Select` is now a native `<select>`, because Radix renders a hidden native select browsers won't report validity on ([primitives#1592](https://github.com/radix-ui/primitives/issues/1592)), leaving it the one required control that could be submitted empty; and the submit button is no longer disabled on invalid input, with a `pattern` on the email field so the browser explains the problem instead of leaving a dead button. Shared blocks pick up small correctness fixes along the way: `Hero` and `HighlightsSection` were keying list items off ids Tina never sets, so every entry shared one `undefined` key; `HighlightsSection` moves its full-bleed band off `w-screen` (which overflows by the scrollbar width) and its column headings from `h4` to `h2` so heading order is valid here and on `/about`; and `renderButtonArrayHelper` now wraps every button type in one uniform cell instead of per-variant margin nudges.


isaaclombardssw:
https://tina-a5d0bjyx7-tinacms.vercel.app/partners#find-a-partner-section

@hveraus @tiagov8 please review and provide feedback

joshbermanssw:
Screenshot 2026-07-15 at 4 22 20 pm

No field valdiation for required fields

kulesy:
image

**Figure: Add to this section ideally above certifications**

joshbermanssw:
Screenshot 2026-07-15 at 4 37 53 pm


$200 fine - em dash

isaaclombardssw:
All addressed in `4925cfd1`:

- @joshbermanssw (validation): required fields now validate natively on submit. Removed the silently-disabled submit button and marked the partner dropdowns required, so empty or invalid fields (including the website URL) show inline feedback.
- @kulesy: added a **Partners** link to the Community nav dropdown, above Certifications.
- @joshbermanssw (em dash): fixed the caption to "Join the herd. Build with TinaCMS." and swapped the en dashes in the team-size options too.

Preview will redeploy shortly. The dropdown validation uses Radix native form support, so worth a quick click-through once it is live.

isaaclombardssw:
https://tina-kmclakjmq-tinacms.vercel.app/partners <-- demo

components/modals/ContactForm.tsx
```diff
@@ -88,7 +189,7 @@ export const ContactForm = () => {
         setFormData(initialFormData);
       } else {
         setMessage({
-          text: 'error',
+          text: '',
```
18-th, line 200: This change effectively disables the logic on :236 since it needs the text to be set to fire

components/modals/ContactForm.tsx
```diff
@@ -212,34 +310,107 @@ export const ContactForm = () => {
...
+        placeholder={companyPlaceholder}
         name="company"
         type="text"
         value={formData.company}
         onChange={handleInputChange}
         disabled={isProcessing}
         className="w-full"
       />
-      <Select
+      {variant === 'partner' && (
```
18-th, line 350: Verified on the preview. It is currently possible to select no options for this field

Image

And the form is happily accepted despite the field being marked as required. Odd UX

components/modals/ContactForm.tsx
```diff
@@ -248,31 +419,33 @@ export const ContactForm = () => {
...
+            Subscribe to the TinaCMS newsletter
+          </span>
+        </label>
+      )}
       <div className="w-full flex justify-end">
         <Button
           type="submit"
           color="orange"
-          disabled={isProcessing || !isValidEmail || !formData.message}
+          disabled={isProcessing || !isValidEmail}
```
18-th: I would expect the 'isValidEmail' guard to be dropped here and enforce the validation on the email field proper. Currently, there is no indication that the email is invalid on the field itself, so it's hard to tell why the submit is disabled.

Image

Invalid email, no message, button disabled with no clear instructions

content/main/partners.json
```diff
@@ -0,0 +1,126 @@
+{
+  "seo": {
+    "title": "TinaCMS Partners – Find an Agency or Become One",
+    "description": "Work with a vetted TinaCMS agency partner to build fast, Git-backed websites with a best-in-class editing experience – or join the program to grow your agency."
```
18-th:
```suggestion
    "description": "Work with a vetted TinaCMS agency partner to build fast, Git-backed websites with a best-in-class editing experience or join the program to grow your agency."
```

content/main/partners.json
```diff
@@ -0,0 +1,126 @@
...
+  },
+  "blocks": [
+    {
+      "blockSettings": {
+        "isHeadingOne": true
+      },
+      "headline": "Build with a TinaCMS partner.",
+      "text": "Find a vetted agency or freelance developer – or become one.",
```
18-th:
```suggestion
      "text": "Find a vetted agency or freelance developer or become one.",
```

components/modals/ContactForm.tsx
isaaclombardssw: ✅ Done

components/modals/ContactForm.tsx
isaaclombardssw: ✅ Done

components/modals/ContactForm.tsx
isaaclombardssw: ✅ Done

content/main/partners.json
isaaclombardssw: ✅ Done

content/main/partners.json
isaaclombardssw: ✅ Done

isaaclombardssw:
@18-th hoping to get a re-review on this. Stripped out dependency on react-select to simplify

components/modals/ContactForm.tsx
isaaclombardssw: Guard now keys off `message.type` rather than `message.text`, so both error paths render.

components/modals/ContactForm.tsx
isaaclombardssw: Ended up going native `<select>`. Radix's Select renders a hidden native select that browsers won't report validity on ([primitives#1592](https://github.com/radix-ui/primitives/issues/1592)), so simplified to native select over fighting with the 'required'.

components/modals/ContactForm.tsx
isaaclombardssw: Guard dropped. Added `pattern` + `title` on the field so it still requires a domain/TLD.

kulesy:
Pushed b72a23e3 with the review fixes: ImageBand's `md:aspect-video` never applied (the named utility lost the cascade to the arbitrary `aspect-[3/2]`) so the band rendered 3:2 at every breakpoint; dropped PartnerGrid's `isSingle` branch, which hid the section heading and made the contact card vanish once a second partner was added; the contact card's copy and email are template fields now, and an explicit `anchorId` replaces the title-derived id that would break the hero link on rename; HighlightsSection's column headings moved h4 to h2 to fix heading order here and on /about (sizing is class-driven, so nothing moves); ContactForm normalises a scheme-less URL on blur rather than letting `type="url"` reject `ssw.com.au`, and requires the agency name for agency applicants; `/api/contact` whitelists `inquiryType`, which is unauthenticated input landing in the email subject. Also refreshed the description, which had drifted from the diff. Lint, types and 82 tests pass, verified on a local dev server.

kulesy (approved):

PR #4656, opened 2026-06-23
Remove formattedDate (we don't use this field anymore)

joshbermanssw:
cc: @isaaclombardssw @adamcogan 

We removed this field from the docs collection - need to remove from the page too


isaaclombardssw (approved):
🚮

isaaclombardssw:
Fix lints

PR #4657, opened 2026-06-23
✨ Add multi-accordion embeddable component to docs

joshbermanssw:
## TL;DR

Ports the `accordionBlock` (multi-accordion) component from `tina-docs` into tina.io as an embeddable MDX component, and uses it to collapse the TinaCloud troubleshooting page's 10 stacked Q&A sections into one accordion.

**Files to review (4, +204 / -105):**

| File | Why |
|---|---|
| `components/tinaMarkdownComponents/docAndBlogComponents.tsx` *(start here)* | New `accordionBlock` renderer + its type in the `Components` generic. |
| `tina/collectionsSchema/docs.tsx` | `accordionBlock` template added to the docs `body` rich-text field. |
| `content/docs/tinacloud/troubleshooting.mdx` | Q&A sections migrated into a single `<accordionBlock>`; duplicate `### N.` sections removed. |
| `tina/tina-lock.json` | Regenerated schema lock for the new template. |

## How

- **Renderer** stacks collapsible items, each with a heading, rich-text body, and optional image; per-item open/close state, kept in sync via `useEffect` when editors add/remove items in the CMS.
- **Schema** exposes `fullWidth` + a list of `accordionItems` (`heading`, `docText` rich-text, `image`).

## Reviewer notes

- **No `isBody` on `docText`.** `isBody: true` on a nested rich-text field stops Tina parsing it into an AST (it serializes as a raw string), so the field mirrors `apiReference`'s `description` instead — that's what makes rich text render inside an item.
- **Grid overflow guard.** Content grid uses `grid-cols-1` (`minmax(0,1fr)`) + `min-w-0 break-words` so long code spans/URLs wrap instead of pushing the card off-screen.
- **YouTube embeds → links.** The two videos in the troubleshooting page became inline `**Video:** [title](url)` links — `<Youtube>` doesn't render inside nested rich-text list items.

## Follow-up

- `ImageAndText` shares the same grid pattern and would hit the same horizontal-overflow bug with long content; not touched here.


components/tinaMarkdownComponents/templateComponents/accordionBlock.tsx
```diff
@@ -0,0 +1,115 @@
+/** biome-ignore-all lint/performance/noImgElement: <TODO> */
+/** biome-ignore-all lint/a11y/noRedundantAlt: <TODO> */
+/** biome-ignore-all lint/correctness/useExhaustiveDependencies: <TODO> */
+/** biome-ignore-all lint/suspicious/noArrayIndexKey: <TODO> */
```
isaaclombardssw: ?

isaaclombardssw (approved):
Some weird comments... but otherwise looks alright.

components/tinaMarkdownComponents/templateComponents/accordionBlock.tsx
joshbermanssw: Good catch — those were placeholder `<TODO>` lint suppressions copied in from the source component. Removed the whole `biome-ignore-all` block in e7628d7b: `noRedundantAlt` and `useExhaustiveDependencies` never actually fired, and the real `noArrayIndexKey` error is now fixed by keying each item on its heading. The only thing left is a non-blocking `noImgElement` warning, which matches how the rest of the codebase uses ``.

PR #4658, opened 2026-06-23
📝 Credit human co-author over bot for docs last-updated

joshbermanssw:
## TL;DR

The docs "Last updated by" widget read `commit.author.name` directly, so any page whose latest commit was authored by `tinacloud-app[bot]` credited the bot instead of the real editor named in the `Co-authored-by:` trailer. This PR substitutes the co-author when the commit author is a bot.

## Why

TinaCloud editorial-workflow commits are authored by the GitHub App (`tinacloud-app[bot]`), with the human recorded as a co-author. For example, `/docs/tinacloud/troubleshooting` currently shows **"Last updated by tinacloud-app[bot]"** when the actual editor was Eli Kent. The metadata route never parsed co-author trailers, so the human was dropped before reaching the UI.

## How

New `resolveDisplayAuthor` substitutes the first `Co-authored-by:` name when the commit author ends in `[bot]` — matching by suffix rather than a literal string so it covers both `tinacloud-app[bot]` and the older `tina-cloud-app[bot]` seen in history. Human-authored commits are returned unchanged (their co-authors, if any, are ignored). The widget now calls this instead of reading `commit.author.name`.

## Reviewer notes

- **Suffix match, not literal.** Catches both bot name spellings + future bots. Tighten to an exact string if you'd rather be strict.
- **Component, not API.** The commit `message` is already in the response, so no API/response-type change was needed; resolver is pure and unit-tested.

## Tests

`components/PageMetadata/authorUtils.test.ts` — 8 Jest cases: trailer parsing (single/multiple/none/undefined) and author resolution (bot→co-author, dashed-bot variant, human unchanged, bot with no co-author falls back to the bot name).

```
Tests: 8 passed, 8 total
```

### DIFF
--- a/components/PageMetadata/authorUtils.test.ts
+++ b/components/PageMetadata/authorUtils.test.ts
@@ -5,7 +5,11 @@ const makeCommit = (authorName: string, message: string): GitHubCommit => ({
   sha: 'abc123',
   commit: {
     author: { name: authorName, email: 'a@example.com', date: '2026-06-01' },
-    committer: { name: 'GitHub', email: 'noreply@github.com', date: '2026-06-01' },
+    committer: {
+      name: 'GitHub',
+      email: 'noreply@github.com',
+      date: '2026-06-01',
+    },
     message,
   },
   author: null,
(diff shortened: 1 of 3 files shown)


isaaclombardssw (approved):
LGTM

PR #4689, opened 2026-07-07
Populate homepage "Recent Videos" from the TinaCMS YouTube channel

### LINKED ISSUE #4688: Homepage "Recent Videos" section goes stale — populate it from the YouTube channel
The "Recent Videos" section on the homepage is hand-authored in `content/main/home.json`. The two video cards only change when someone manually pastes new YouTube embed URLs, so they drift out of date and don't reflect the channel's latest uploads.

**Goal:** the two cards should auto-populate with the most recent uploads from the TinaCMS YouTube channel (https://www.youtube.com/@TinaCMS), with no manual editing.

**Approach:**
- Fetch the latest uploads from the channel's public RSS feed (`youtube.com/feeds/videos.xml`) — no API key or quota needed.
- Fetch server-side with hourly ISR.
- Keep the editorially chosen "Featured Video" manual, and exclude it from the list so it never shows twice.
- Keep the curated `youtubeVideos` in `home.json` as a fallback if the feed can't be fetched (offline/static-export build or network flake), so the section never renders empty.

**Out of scope / follow-ups:**
- Filtering out Shorts (the RSS feed doesn't expose video type/duration).
- Per-presenter SSW People links — the feed only gives the channel, so the presenter name is parsed from the `Title | Name` convention and links to the channel.

isaaclombardssw:
Closes #4688

The homepage "Recent Videos" cards were hand-authored in `content/main/home.json`, so they went stale until someone manually pasted new embed URLs. This makes the two cards below the featured video auto-populate from the TinaCMS YouTube channel.

## Approach
- `utils/youtube/getChannelVideos.ts` lists the channel's uploads playlist via the **YouTube Data API v3** (`playlistItems.list`, ~1 quota unit/call against a 10k/day budget) using `YOUTUBE_API_KEY`.
- `app/api/recent-videos` proxies it so the block can fetch **client-side** without exposing the key. It never caches an empty response, so a transient failure can't freeze the CDN.
- `RecentPostsBlock` fetches the route on mount, shows a **skeleton loader**, fades the cards in (`animate-row-in`), excludes the editorially chosen featured video, and falls back to the curated `youtubeVideos` in `home.json` on empty/error.
- The "Featured Video" stays editor-controlled. Presenter name is parsed from the `Title | Name` convention; author links to the channel.

## Why the Data API (not RSS)
The first cut used the public RSS feed (`youtube.com/feeds/videos.xml`). It worked from a browser/dev machine but **YouTube rate-limits / blocks the RSS endpoint from datacenter IPs**, so on Vercel it returned empty most of the time and the block silently showed the curated fallback ([YT 429 background](https://decodo.com/blog/youtube-error-429)). The Data API is key-authenticated and built for server use, so it's reliable from serverless.

## ⚠️ Deploy requirement
`YOUTUBE_API_KEY` must be set in the Vercel project env (Preview + Production) — a YouTube Data API v3 key from Google Cloud. Without it the block gracefully falls back to the curated videos.

## Notes / follow-ups
- No TinaCMS schema change, so no `tina-lock.json` churn. The `youtubeVideos` field is now effectively a fallback.
- Shorts are not filtered.

## Testing
- `pnpm type-check` and `pnpm lint` clean.
- Verified skeleton, fade-in, featured-exclusion, and curated fallback on the preview. Live Data API path pending the env key.


utils/youtube/getChannelVideos.ts
```diff
@@ -0,0 +1,101 @@
...
+  channelId: string = TINACMS_YOUTUBE_CHANNEL_ID,
+  excludeVideoIds: string[] = [],
+): Promise<ChannelVideo[]> {
+  try {
+    const res = await fetch(
+      `https://www.youtube.com/feeds/videos.xml?channel_id=${channelId}`,
+      { next: { revalidate: REVALIDATE_SECONDS } },
+    );
+    if (!res.ok) {
+      return [];
```
joshbermanssw: Why empty return on not ok?

utils/youtube/getChannelVideos.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
```
joshbermanssw, line 65: also + a `catch { ... }`

utils/youtube/getChannelVideos.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    const xml = await res.text();
+    // Drop everything before the first <entry> (channel-level metadata).
+    const entries = xml.split('<entry>').slice(1);
+    const videos: ChannelVideo[] = [];
+
+    for (const entry of entries) {
```
joshbermanssw: Does this go through ALL videos in the channel? Can we jsut limit the loop to 3?

utils/youtube/getChannelVideos.ts
```diff
@@ -0,0 +1,101 @@
...
+      });
+
+      if (videos.length >= count) {
+        break;
+      }
+    }
+
+    return videos;
+  } catch {
+    // Network flake / feed unavailable — caller falls back to curated content.
```
joshbermanssw: can we have some sort of response message on err pls

app/[...slug]/page.tsx
```diff
@@ -55,11 +57,30 @@ export default async function Page({ params }: PageProps) {
     const res = await queryFunction({
       relativePath: relativePath,
     });
+
+    // Only pages that actually render the "Recent Videos" block need the live
```
Comment (joshbermanssw) [inline app/[...slug]/page.tsx]: Can we request the lagestVideos at the component level rather than page level/

joshbermanssw (changes requested):

utils/youtube/getChannelVideos.ts
isaaclombardssw: Warns now

utils/youtube/getChannelVideos.ts
isaaclombardssw: Warns now

utils/youtube/getChannelVideos.ts
isaaclombardssw: Limited to 3 :)

utils/youtube/getChannelVideos.ts
isaaclombardssw: Added — the `catch` now `console.warn`s with the underlying error message before falling back.

app/[...slug]/page.tsx
Comment (isaaclombardssw) [reply app/[...slug]/page.tsx]: It has to stay server-side, unfortunately: the block renders inside the client `useTina` tree so it can't be an async server component, and YouTube's RSS feed sends no `Access-Control-Allow-Origin` header, so a browser fetch from the component would be CORS-blocked (an API-route proxy would work but loses SSR — videos would pop in with layout shift). Server-fetching also keeps them in the initial HTML.

What I did change: pulled the block-detection + featured-exclusion out of the catch-all route into a `getRecentVideos()` helper co-located with the block, so `page.tsx` is now a one-liner and the logic lives next to the component that uses it. Happy to switch to the API-route approach if you'd prefer it despite the SSR tradeoff.

utils/youtube/getChannelVideos.ts
```diff
@@ -0,0 +1,111 @@
...
+// key and has no quota, so the homepage "Recent Videos" section can stay
+// current without anyone hand-editing content. Server-only.
+
+// TinaCMS channel: https://www.youtube.com/@TinaCMS
+export const TINACMS_YOUTUBE_CHANNEL_ID = 'UCUvqCjr8Xq_IRMDcuJrqIXA';
+const TINACMS_YOUTUBE_CHANNEL_URL = 'https://www.youtube.com/@TinaCMS';
+
+// Re-fetch the feed at most once an hour (ISR). Uploads are infrequent, so a
+// longer window keeps build/runtime fetches cheap without going stale.
+const REVALIDATE_SECONDS = 3600;
```
joshbermanssw: as per our convo - moving to daily revalidation

app/[...slug]/page.tsx
Comment (joshbermanssw) [reply app/[...slug]/page.tsx]: as per our conversation we'll move to csr and fetch on the component as the ssr can reduce time to first load for the entire homepage (especially if for some reason YT dependency is slow), Isaac mentioned he'll add a skeleton loader.

utils/youtube/getChannelVideos.ts
isaaclombardssw: Done — `REVALIDATE_SECONDS` is now 86400 (1 day), and the new proxy route caches for a day to match.

app/[...slug]/page.tsx
Comment (isaaclombardssw) [reply app/[...slug]/page.tsx]: Moved to CSR. Added `app/api/recent-videos` (a thin proxy — needed because YouTube's feed sends no CORS header, so the browser can't hit it directly), and the block now fetches that on mount with a skeleton loader while it loads, falling back to the curated videos on empty/error. Server-side threading is reverted. Ready for your skeleton styling if you want to tweak it.

.env.example
```diff
@@ -18,4 +18,7 @@ GISCUS_CATEGORY_ID=***
 GISCUS_THEME_URL=***
 SENDGRID_API_KEY=***
 SENDGRID_FROM_EMAIL=***
-CONTACT_FORM_TO=***
\ No newline at end of file
+CONTACT_FORM_TO=***
+# YouTube Data API v3 key — powers the homepage "Recent Videos" block.
```
joshbermanssw: rm comments

components/blocks/RecentPosts/RecentPosts.tsx
```diff
@@ -73,8 +84,59 @@ const VideoCard = ({
...
+  const curatedVideos: VideoCardProps[] = data?.youtubeVideos ?? [];
+
+  // Fetch the latest uploads client-side via our proxy route so the YouTube
+  // dependency never blocks the homepage's first paint.
+  const [fetchedVideos, setFetchedVideos] = useState<VideoCardProps[] | null>(
+    null,
+  );
+  const [fetchFailed, setFetchFailed] = useState(false);
+
+  useEffect(() => {
```
joshbermanssw, line 106: can we turn this back into a server component and await the data req?

components/blocks/RecentPosts/RecentPosts.tsx
```diff
@@ -73,8 +84,59 @@ const VideoCard = ({
...
+
+  const isLoading = fetchedVideos === null && !fetchFailed;
+
+  // Drop the editorially chosen featured video so it never appears twice.
+  const featuredId = extractYouTubeId(featuredPost?.url);
+  const liveVideos = (fetchedVideos ?? [])
+    .filter((video) => extractYouTubeId(video.embedUrl) !== featuredId)
+    .slice(0, RECENT_VIDEO_COUNT);
+
+  // Prefer the live feed; fall back to the curated videos in content when the
```
joshbermanssw: can we add a note in the description of the component schema telling the editor that this is the case

utils/youtube/getChannelVideos.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+  return [title, 'TinaCMS'];
+};
+
+export async function getChannelVideos(
+  count = 2,
+  channelId: string = TINACMS_YOUTUBE_CHANNEL_ID,
+  excludeVideoIds: string[] = [],
+): Promise<ChannelVideo[]> {
+  const apiKey = process.env.YOUTUBE_API_KEY;
```
joshbermanssw, line 42: is this api key in Keeper?

utils/youtube/getChannelVideos.ts
```diff
@@ -0,0 +1,130 @@
...
+const decodeEntities = (text: string): string =>
+  text
+    .replace(/&amp;/g, '&')
+    .replace(/&lt;/g, '<')
+    .replace(/&gt;/g, '>')
+    .replace(/&quot;/g, '"')
+    .replace(/&#0?39;/g, "'")
+    .replace(/&#(\d+);/g, (_, code) => String.fromCharCode(Number(code)));
+
+// SSW video titles follow a "Video Title | Presenter Name" convention. Split
```
joshbermanssw: I'd avoid hard-coding this convention as it isn't "guaranteed"

utils/youtube/getChannelVideos.ts
```diff
@@ -0,0 +1,130 @@
...
+    const author = parts.pop().trim();
+    if (author) {
+      return [parts.join('|').trim(), author];
+    }
+  }
+  return [title, 'TinaCMS'];
+};
+
+export async function getChannelVideos(
+  count = 2,
```
joshbermanssw: can we mvoe the random num to a variable at the top just calling it sm like DEFAULT_CHANNEL_VIDEO_COUNT or similar?

joshbermanssw (changes requested):

.env.example
isaaclombardssw: ✅ Done — removed.

components/blocks/RecentPosts/RecentPosts.tsx
isaaclombardssw: ✅ Added a note to the `youtubeVideos` field description explaining it is a fallback used only when the live feed can't be loaded.

utils/youtube/getChannelVideos.ts
isaaclombardssw: ✅ Done — extracted to `DEFAULT_CHANNEL_VIDEO_COUNT`.

utils/youtube/getChannelVideos.ts
isaaclombardssw: Yes, the key is in Keeper.

utils/youtube/getChannelVideos.ts
isaaclombardssw: Kept the parsing, but it already degrades safely: when a title has no `|` (or an empty presenter), it falls back to the channel name rather than guessing. Enforcing the `Title | Presenter` convention editorially so the split stays reliable.

components/blocks/RecentPosts/RecentPosts.tsx
isaaclombardssw: Kept it as CSR. The proxy route caches the YouTube response at the edge, but a server component would still put `getChannelVideos` on the homepage's critical render path, so a slow/failing YouTube call would delay first paint of the whole homepage. That's the risk the earlier thread moved to CSR + skeleton to avoid, and the proxy doesn't remove it for SSR. Happy to revisit if you'd prefer SSR despite that.

joshbermanssw (approved):

PR #4697, opened 2026-07-10
docs: fix Cloudflare Workers guide (SESSION KV, runtime vars) + add CLI path

isaaclombardssw:
Reworks the Cloudflare Workers deployment guide for the Astro starter after running a real deploy end to end and verifying each claim against the built worker and live Cloudflare/Astro docs.

## What was wrong

**Runtime environment variables (incorrect).** The guide told readers to re-add `PUBLIC_TINA_CLIENT_ID` and `TINA_TOKEN` as runtime variables because "the island route reads them at request time." It doesn't. `PUBLIC_TINA_CLIENT_ID` is inlined at build (standard Astro/Vite behaviour) and `TINA_TOKEN` is baked into the generated Tina client as a string literal. Greps of the built worker show no runtime `process.env` read of either. The runtime copies are inert, so I removed that guidance and the matching troubleshooting line.

**SESSION KV namespace (missing, and a real failure).** The `@astrojs/cloudflare` adapter injects a `SESSION` KV binding with no id, which the guide never mentioned. The first deploy provisions it, but Cloudflare does not write the id back to the repo, so the next git-based deploy tries to create it again and fails with `code: 10014` (namespace already exists). Because editor saves trigger a redeploy, this bites on the second save. Added a section documenting it with the fix: pin the namespace id in `wrangler.jsonc`. I verified the pinned id flows through the adapter's generated config and that a redeploy then reuses the namespace with no provisioning step or collision.

Refs: withastro/astro#15802, emdash-cms/emdash#437.

**Build output field (inapplicable).** Dropped "Build output: `dist`" — a wrangler-deployed Worker gets its assets from the adapter config, not a dashboard directory. There is no such field in the Workers Builds UI.

## What was added

**A CLI deploy path.** The guide only covered Cloudflare's git-based builds. A local `pnpm build` selects the Node adapter (because `WORKERS_CI` is only set inside Cloudflare's builds), so a from-your-machine deploy needs `DEPLOY_ADAPTER=cloudflare pnpm build` before `npx wrangler deploy`. Documented as a second path.

## Also

Rewrote the page for concision (led with the commands, cut the hand-holding) and removed em dashes from both `cloudflare-workers.mdx` and `cloudflare-pages.mdx`.

Verified against @astrojs/cloudflare ^13.7.0, astro ^6.4.4, @tinacms/astro 0.5.1, wrangler 4.110.


joshbermanssw:
Nice cleanup, especially the `SESSION` KV pinning and confirming the runtime vars are inert.

One gap this guide (and the starter) still has for Cloudflare: **the Tina branch is silently pinned to `main` on any Cloudflare deploy.**

`tina/config.ts` in `tina-astro-starter` resolves the branch like this:

```js
const branch =
  process.env.GITHUB_BRANCH ||
  process.env.VERCEL_GIT_COMMIT_REF ||
  process.env.HEAD ||
  "main";
```

Cloudflare sets none of those. Workers Builds exposes the branch as `WORKERS_CI_BRANCH` (and Pages as `CF_PAGES_BRANCH`), neither of which is read here, so `branch` always falls back to `"main"`. Vercel is fine because `VERCEL_GIT_COMMIT_REF` is handled, which is why this is easy to miss.

Impact: a production-from-`main` deploy works, but a non-`main` production branch or a preview deploy will build the site from one branch while TinaCMS reads/writes content on `main`. Since we're making this the flagship Astro-on-Cloudflare path, that mismatch is worth closing.

Best fix is in the starter's `config.ts` (add `process.env.WORKERS_CI_BRANCH || process.env.CF_PAGES_BRANCH` to the chain). At minimum, this guide could carry a one-line note that on Cloudflare the Tina branch defaults to `main` unless you set the branch env var explicitly.

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
```diff
@@ -1,31 +1,27 @@
...
+Deploy to Cloudflare Workers when your site needs a server at request time: SSR, API routes, redirects, or the Astro starter's visual-editing route. Every push to your production branch rebuilds and redeploys.
 
 <WarningCallout
   body={<>
-    **Astro users:** deploy to Workers, not [Cloudflare Pages](/docs/tinacloud/deployment-options/cloudflare-pages). The `@astrojs/cloudflare` adapter [no longer supports Cloudflare Pages](https://docs.astro.build/en/guides/integrations-guide/cloudflare/) — it targets Workers. The starter's visual-editing route (`src/pages/tina-island/[name].ts`) is server-rendered (`prerender = false`), so a static-only Pages upload would break click-to-edit at request time.
+    **Astro users:** deploy to Workers, not [Cloudflare Pages](/docs/tinacloud/deployment-options/cloudflare-pages). The `@astrojs/cloudflare` adapter [no longer supports Pages](https://docs.astro.build/en/guides/integrations-guide/cloudflare/), and the starter's visual-editing route (`src/pages/tina-island/[name].ts`) is server-rendered, so a static Pages upload breaks click-to-edit.
```
joshbermanssw:
```suggestion

```

I'd remove the entire Warning Callout - the users are on Cloudflare Workers page already

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
```diff
@@ -1,31 +1,27 @@
...
   </>}
 />
 
 ## Astro (the `tina-astro-starter`)
```
joshbermanssw:
```suggestion
## Astro
```

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
```diff
@@ -1,31 +1,27 @@
...
 
-This walks the [Astro starter](https://github.com/tinacms/tina-astro-starter) from a clean checkout to a live Workers deploy. Push your project to a GitHub repository first.
-
-### 1. Adapter and `wrangler.jsonc` (already in the starter)
+Push your project to GitHub, then deploy it with Cloudflare's git-based builds or straight from your machine. Both use the config the starter already ships.
```
joshbermanssw:
```suggestion
```

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
```diff
@@ -1,31 +1,27 @@
...
 
-Nothing to configure — the starter ships this for you:
+### What the starter already sets up
 
-* `astro.config.mjs` **auto-detects the host**. In Cloudflare's git-based Workers Builds, the `WORKERS_CI` environment variable is set automatically, which selects the `@astrojs/cloudflare` adapter. You don't run `npx astro add cloudflare`, and you don't edit the config.
-* A root **`wrangler.jsonc`** enables `nodejs_compat`. The visual-editing route uses `node:async_hooks` (`AsyncLocalStorage`); without this flag the island route fails at runtime:
+* `astro.config.mjs` picks the adapter from the host. Cloudflare's builds set `WORKERS_CI`, which selects `@astrojs/cloudflare`, so you never run `npx astro add cloudflare` or edit the config.
```
joshbermanssw:
```suggestion
* `astro.config.mjs` detects the adapter from the host. Cloudflare's builds set `WORKERS_CI`, which selects `@astrojs/cloudflare`.
```

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
```diff
@@ -1,31 +1,27 @@
...
+* A root `wrangler.jsonc` turns on `nodejs_compat`, which the visual-editing route needs for `node:async_hooks`:
```
joshbermanssw:
```suggestion
* A root `wrangler.jsonc` enables `nodejs_compat`, which the visual-editing route needs:
```

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
```diff
@@ -1,31 +1,27 @@
...
 
 ```jsonc
```
joshbermanssw, line 20: does the user need to see this codeblock?

isaaclombardssw:
Good catch, confirmed. `WORKERS_CI_BRANCH` (Workers Builds) and `CF_PAGES_BRANCH` (Pages) are the right vars, and the current `config.ts` chain reads neither, so it falls back to `main` on Cloudflare.

Fixed both ways:
- Root fix in the starter: tinacms/tina-astro-starter#79 adds `WORKERS_CI_BRANCH` and `CF_PAGES_BRANCH` to the branch chain.
- Added an "Editing branch" note to this guide for anyone on an older `config.ts`.

`WORKERS_CI_BRANCH` verified against the Workers Builds configuration docs.

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
```diff
@@ -35,63 +31,85 @@ Nothing to configure — the starter ships this for you:
...
+> Adding TinaCMS to an existing Astro site? Run `npx astro add cloudflare`, then add a root `wrangler.jsonc` with `compatibility_flags: ["nodejs_compat"]` and a recent `compatibility_date`.
 
-> **Adding TinaCMS to an existing Astro site?** Run `npx astro add cloudflare`, then create a `wrangler.jsonc` at the repo root with `compatibility_flags: ["nodejs_compat"]` and a recent `compatibility_date`.
+### Deploy with git-based builds
 
-### 2. Create the Worker
+In the Cloudflare dashboard, go to **Workers & Pages** | **Create** | **Workers** | **Import a repository**, connect GitHub, and pick your repo. Set:
 
-In the Cloudflare dashboard go to **Workers & Pages** | **Create** | **Workers** | **Import a repository**, connect your GitHub account, and select your repo.
+* **Build command:** `pnpm run build` (the starter's script runs `tinacms build --content=local -c "astro build"`)
```
joshbermanssw:
```suggestion
* **Build command:** `pnpm run build`
```

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
```diff
@@ -35,63 +31,85 @@ Nothing to configure — the starter ships this for you:
...
+* **Deploy command:** `npx wrangler deploy`
 
-Set the build under **Build configuration**:
+There's no build-output field to set. A Worker gets its static assets from the adapter's generated config, not a dashboard directory, and Cloudflare authenticates the deploy for you.
```
joshbermanssw, line 37:
```suggestion

```

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
```diff
@@ -35,63 +31,85 @@ Nothing to configure — the starter ships this for you:
...
 
-* **Build command:** `pnpm build` — the starter's `build` script is `tinacms build --content=local -c "astro build"`, which indexes your committed content, then runs `astro build`.
-* **Deploy command:** `npx wrangler deploy`
-* **Build output:** `dist` (Astro's default; you don't need to change this).
+### Or deploy from your machine
```
joshbermanssw: Does the clanker mean the CLI?

```suggestion
### Or deploy via the CloudFlare CLI 
```

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
```diff
@@ -35,63 +31,85 @@ Nothing to configure — the starter ships this for you:
...
 
-`DEPLOY_ADAPTER` is **not** needed here — `WORKERS_CI` is detected automatically. You only set `DEPLOY_ADAPTER=cloudflare` for a manual deploy where no host env var applies (e.g. running `npx wrangler deploy` from your own machine).
+Put your TinaCloud credentials in a local `.env` (see below), then build and deploy:
```
joshbermanssw:
```suggestion
Put your TinaCloud credentials in a local `.env`, then build and deploy:
```

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
```diff
@@ -35,63 +31,85 @@ Nothing to configure — the starter ships this for you:
...
 
-<WarningCallout
-  body={<>
-    Workers keep **build** variables and **runtime** variables separate. The island route reads your TinaCloud client ID and token at request time, so add `PUBLIC_TINA_CLIENT_ID` and `TINA_TOKEN` again under **Settings** | **Variables and Secrets** (not just the build section). `SITE_URL` is only used at build time.
-  </>}
-/>
+`tinacms build` needs your TinaCloud credentials at build time. Add them wherever you build: for git-based builds, under the Worker's **Settings** | **Build** | **Variables and Secrets**; for a CLI deploy, in your local `.env`. Grab both credentials from your [TinaCloud project](https://app.tina.io).
 
-### 4. Verify visual editing works at request time
+* `PUBLIC_TINA_CLIENT_ID` is your client ID. Astro uses the `PUBLIC_` prefix, not Next.js's `NEXT_PUBLIC_`.
```
joshbermanssw, line 54:
```suggestion
* `PUBLIC_TINA_CLIENT_ID` is your client ID. 
```

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
```diff
@@ -35,63 +31,85 @@ Nothing to configure — the starter ships this for you:
...
+* `TINA_TOKEN` is a content token.
+* `SITE_URL` is your production URL, e.g. `https://your-worker.your-account.workers.dev`. Workers injects no deploy URL, so without it your sitemap, RSS, and OpenGraph tags fall back to `localhost`.
```
joshbermanssw, line 56:
```suggestion
* `SITE_URL` is your production URL, e.g. `https://your-worker.your-account.workers.dev`. 
```

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
```diff
@@ -35,63 +31,85 @@ Nothing to configure — the starter ships this for you:
...
 
-After the first deploy:
+These are build-time only. `tinacms build` bakes the client ID and token into the generated client, so there's nothing to add again as a runtime variable.
 
-1. Open `/admin` on your Worker's URL, edit a post, and save. The save commits to GitHub, which triggers a rebuild and redeploy.
-2. Open a content page and confirm **click-to-edit** works: clicking an editable region highlights it and opens its form in the sidebar. This exercises the on-demand `/tina-island/*` route running inside the Worker — if it errors, check that `nodejs_compat` is set and that your runtime variables are present.
+### Pin the SESSION KV namespace
```
joshbermanssw, line 78: As a user reading these docs, a. what is it, and b. why do i need to pin it?

joshbermanssw (dismissed):

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
18-th: There are multiple adaptors from my knowledge. That wording is correct

isaaclombardssw:
Read the above, a bunch of things are assuming they're using the astro starter so I'll need to make it more generic

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
18-th: That's useful context, I wouldn't remove it

isaaclombardssw:
Reworked the Astro section to be starter-agnostic with "Using the starter?" asides, which also resolves what your suggestions surfaced: `DEPLOY_ADAPTER` and the `WORKERS_CI` auto-detection are logic in the starter's `astro.config.mjs`, not stock Astro, so they'd be wrong for a hand-rolled app under a plain "## Astro" heading. They now live in asides; the main steps (`npx astro add cloudflare`, `adapter: cloudflare()`, a plain `pnpm run build && wrangler deploy`) are correct for any Astro app.

Mapping to your comments:
- Pages warning callout: removed entirely (the empty shell too).
- `## Astro`, `pnpm run build`, "deploy from the Wrangler CLI", and the bullet/env trims: applied.
- wrangler.jsonc codeblock: kept, since a non-starter reader has to create it. Starter users get an aside saying it is already there.
- SESSION KV ("what is it, why pin it"): added a line that the adapter injects it for Astro's session store, plus why pinning avoids the `10014` collision on redeploys.

One I did not cut: the `SITE_URL` "Workers injects no deploy URL" clause. It is a genuine Cloudflare-only gotcha (Vercel/Netlify/Pages set a URL var, Workers does not), so without it canonicals silently become `localhost`. Happy to drop it if you would rather keep the list terse.

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
isaaclombardssw: Agreed, kept it. Since the guide is now starter-independent, this wording (the host picking among the supported adapters) lives in the "Using the starter?" aside, because that host auto-detection is the starter's astro.config logic rather than stock Astro. The generic path just uses `npx astro add cloudflare`.

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
isaaclombardssw: Kept it. Left the SITE_URL rationale in as-is.

isaaclombardssw: Kept it, left the SITE_URL rationale in as-is.

content/docs/tinacloud/deployment-options/cloudflare-workers.mdx
isaaclombardssw: yep they need to create that file if they're not using the starter

isaaclombardssw:
Actioned fixes as above.

joshbermanssw:
Duplicate 'Using the starter' boxes

Screenshot 2026-07-14 at 9 39 33 am

isaaclombardssw:
Reworded rather than merged. The two boxes cover different workflows and don't collapse cleanly into one:

- **Git-based builds:** Cloudflare sets `WORKERS_CI`, so `astro.config.mjs` auto-selects `@astrojs/cloudflare`. Nothing to do.
- **Local Wrangler CLI:** no CI var is set, so the config falls back to the Node adapter. You have to force it with `DEPLOY_ADAPTER=cloudflare`.

A shared box would present `DEPLOY_ADAPTER` as a required step for both paths, but git builds don't need it — auto-detection already picks Cloudflare. What made the two look duplicated was both opening with the same "auto-detects the host" line, so I trimmed the CLI box to lead with its consequence instead:

> **Using the starter?** On your own machine no CI variable is set, so the build falls back to the Node adapter. Force Cloudflare: `DEPLOY_ADAPTER=cloudflare pnpm run build`.

Pushed in d61ee4fd.

isaaclombardssw:
Wranger CLI and Git-based are 2 alternative workflows so don't agree with consolidating the boxes (see above)

joshbermanssw:
> Wranger CLI and Git-based are 2 alternative workflows so don't agree with consolidating the boxes (see above)

a. The wording of both of these boxes are confusing because they start with "Using the starter?" and "Using the Astro starter?" 

b. I'd remove the first box - it doesnt help a user in their steps

Screenshot 2026-07-14 at 10 22 22 am

isaaclombardssw:
I feel calling out the starter here is appropriate, it also tries to direct people into using it.

Any opinion @18-th?

18-th:
I feel like the astro starter section is useful. It's a fairly specific setup, and it's useful to call out that less work is needed when using a starter, I would keep it.

18-th (approved):
LGTM

PR #4743, opened 2026-07-29
🐛 Bump tinacms to 3.11.0 to restore the rich-text link popover

joshbermanssw:
## TL;DR

Adding or editing a link in the rich-text editor does nothing visible on `main` — the popover opens but renders behind the sidebar form fields. This bumps `tinacms` to 3.11.0, which contains the fix.

## Reviewer notes

- **The CLI and app bumps are load-bearing, not housekeeping.** `@tinacms/cli@2.5.3` and `@tinacms/app@2.5.8` exact-pin `tinacms: 3.10.0`, so bumping `tinacms` alone leaves the admin build resolving a nested 3.10.0 that still has the bug. 2.5.6 / 2.5.10 use `^3.11.0` and dedupe.
- **Lockfile shrinks by ~180 lines.** That is the duplicate nested `tinacms` tree collapsing, not dependencies being dropped.

## Links

- [tinacms#7250](https://github.com/tinacms/tinacms/issues/7250) — the bug
- [tinacms#7251](https://github.com/tinacms/tinacms/pull/7251) — the fix


18-th (approved):
LGTM

PR #4749, opened 2026-07-30
Fix YouTube Error 153 by declaring a referrer policy on embeds

### LINKED ISSUE #4748: 🐛 /docs/vibe-coding - YouTube "Error 153" on Adam's ipad
### 📧 Source email

> From: Adam Cogan [SSW] <AdamCogan@ssw.com.au>
> Date: Thursday, 30 July 2026 at 10:00 am
> To: Hark Singh [SSW] <HarkSingh@ssw.com.au>; Tiago Araujo [SSW] <TiagoAraujo@ssw.com.au>
> Subject: Vibe Coding with TinaCMS | TinaCMS
>
> Hark
> https://tina.io/docs/vibe-coding
> 1. Missing image
>
> Tiago
> 1. Is this on CodeAuditor

Cc: @adamcogan @tiagov8

### 🐛 Describe the Bug

1 media problems on https://tina.io/docs/vibe-coding:

1. **YouTube embed shows "Error 153" on iPad.** Seen on @adamcogan's iPad. The same embed plays fine on desktop Chrome, so it looks device specific and needs investigating.

 
Image

**Figure: From Adam's email: Error 153**

### 🧑‍💻 Reproduction Steps

1. Go to https://tina.io/docs/vibe-coding
2. On an iPad, play the YouTube embed at the top of the page. It shows "Error 153"

### 🎯 Expected Behaviour

The YouTube embeds play on iPad the same way they do on desktop.

### 💼 Tasks

- [x] Reproduce and fix "Error 153" on the iPad YouTube embed
- [ ] @tiagov8 - check whether CodeAuditor already flags broken images like this

### ✅ Acceptance Criteria

- [x] YouTube embeds on the page play on iPad

### ➕ More Information

Checked on desktop Chrome: the YouTube embed is fine there, the images are broken.

0xharkirat:
Fixes #4748

## The pain

On an iPad, the video at the top of [/docs/vibe-coding](https://tina.io/docs/vibe-coding) does not play.
You get a dark box reading "Error 153 - Video player configuration error".
The box has no thumbnail in it, so it looks like a broken image.
The same page plays in desktop Chrome, so the iPad looked like the cause.

## Investigation

Error 153 comes from YouTube, not from us.
Its player refuses to configure itself when the request for the embed arrives with no `Referer` header.
So the question was not what the iPad does differently, it was who drops that header.

I loaded the live page in Chrome and in WebKit.
For each one I recorded the `Referer` sent to `youtube.com/embed`, then read the error text from inside the iframe:

| Browser | `Referer` sent | Result |
| --- | --- | --- |
| Chrome desktop | `https://tina.io/` | plays |
| WebKit, iPad viewport | `https://tina.io/` | plays |
| WebKit, iPad viewport, referrer withheld | none | Error 153 |
| Chrome desktop, referrer withheld | none | Error 153 |

Desktop Chrome fails the same way once the referrer is withheld, so this is not an iPad bug.
Our embeds never declared a referrer policy, which left the header to whatever context loaded the page.
Browsers send it by default.
iOS in-app WebViews, which is what Outlook and Teams open a tapped link in, do not.
The bug report came from a link in an email, which fits.

The site is not the one dropping it: there is no `Referrer-Policy` header in `next.config.js`, `vercel.json`, or `middleware.ts`.
That leaves the iframe as the only place we can defend against a context that withholds it.

## The solution

Set `referrerPolicy="strict-origin-when-cross-origin"` on the iframe.
An attribute on the iframe overrides the surrounding context, so the header goes out even inside a WebView that would otherwise drop it.
That value is already the default in Chrome, Firefox, and Safari, so a normal visitor's header does not change.
Every embed URL is `https`, so there is no downgrade case.

Applied to the 4 components that render a YouTube embed, plus 2 places carrying the same gap:

- the generic `Iframe` rich-text component, which an editor can point at a YouTube embed URL
- the Astro sample in `content/docs/contextual-editing/astro.mdx`, which taught readers to build the same broken component. Only the English source changes here, because `content/docs-zh/` is regenerated by the auto-translate workflow.

`components/youtubeEmbedReferrerPolicy.test.ts` fails if a YouTube iframe under `app/` or `components/` drops the attribute.
Strip it from one embed and the test names that file.

Checks on this branch: `npx jest` 82 passed in 15 suites, `npx tsc` exit 0, `npx biome ci .` exit 0.
`pnpm build` needs TinaCloud credentials, so CI covers that one.

## Suggested review order

1. `components/tinaMarkdownComponents/docAndBlogComponents.tsx` - the `Youtube` component from the bug report, and the generic `Iframe` component with the same gap.
2. `components/blocks/VideoEmbed/LiteYouTube.tsx` - shared by `RecentPosts` and `VideoDisplay`, so 1 attribute covers 3 call sites.
3. `components/blocks/Features/Features.tsx` and `components/blocks/Media/MediaComponent.tsx` - the last 2 renderers, same line in each.
4. `content/docs/contextual-editing/astro.mdx` - the docs sample that shipped the bug to readers.
5. `components/youtubeEmbedReferrerPolicy.test.ts` - the guard, which reads better once you have seen the embeds it protects.

## Reproduce it

You do not need an iPad.
Withhold the referrer with the Playwright MCP server and read what the player says.
Pass this to `browser_run_code_unsafe`:

```js
async (page) => {
  await page.route('**/docs/vibe-coding', async (route) => {
    const res = await route.fetch();
    const headers = { ...res.headers(), 'referrer-policy': 'no-referrer' };
    delete headers['content-encoding'];
    delete headers['content-length'];
    await route.fulfill({ status: res.status(), headers, body: await res.text() });
  });
  await page.goto('https://tina.io/docs/vibe-coding');
  const frame = page.frameLocator('iframe[src*="buaXi8Hh420"]');
  return await frame.locator('body').innerText({ timeout: 15000 });
}
```

Against production it returns `Watch video on YouTube / Error 153 / Video player configuration error`.

For the fixed version, run `pnpm dev` on this branch and point the same call at `http://localhost:3000/docs/vibe-coding`.
The referrer is still withheld, but the component carries the attribute, so you get the player back: `Vibe blogging with GitHub Copilot & TinaCMS | Hark Singh`.

### DIFF
--- a/components/blocks/Features/Features.tsx
+++ b/components/blocks/Features/Features.tsx
@@ -339,6 +339,7 @@ const YouTubeEmbed = ({ videoId }: { videoId: string }) => (
       src={`https://www.youtube.com/embed/${videoId}?autoplay=1`}
       title="YouTube video player"
       frameBorder="0"
+      referrerPolicy="strict-origin-when-cross-origin"
       allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share"
       allowFullScreen
     ></iframe>
--- a/components/blocks/Media/MediaComponent.tsx
+++ b/components/blocks/Media/MediaComponent.tsx
@@ -16,6 +16,7 @@ const VideoGridComponent = ({ data }) => {
             src={media.embedUrl}
             title="YouTube video player"
             className="object-cover rounded-lg"
+            referrerPolicy="strict-origin-when-cross-origin"
             allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
             allowFullScreen
           />
--- a/components/blocks/VideoEmbed/LiteYouTube.tsx
+++ b/components/blocks/VideoEmbed/LiteYouTube.tsx
@@ -49,6 +49,7 @@ export const LiteYouTube = ({
           title={title}
           className="absolute inset-0 w-full h-full rounded-lg shadow-lg"
           frameBorder="0"
+          referrerPolicy="strict-origin-when-cross-origin"
           allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture"
           allowFullScreen
         />
--- a/components/tinaMarkdownComponents/docAndBlogComponents.tsx
+++ b/components/tinaMarkdownComponents/docAndBlogComponents.tsx
@@ -286,6 +286,7 @@ export const docAndBlogComponents: Components<{
           height={`${height}px`}
           src={iframeSrc}
           title="Iframe"
+          referrerPolicy="strict-origin-when-cross-origin"
         />
       </div>
     );
@@ -558,6 +559,7 @@ export const docAndBlogComponents: Components<{
           className="absolute top-0 left-0 w-full h-full rounded-md"
           src={embedSrc}
           title="YouTube video player"
+          referrerPolicy="strict-origin-when-cross-origin"
           allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
           allowFullScreen={true}
         ></iframe>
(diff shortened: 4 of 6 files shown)


isaaclombardssw (approved):
✅

PR #4775, opened 2026-08-17
docs: document the Auth.js interop workaround for non-Next.js backends

### LINKED ISSUE #4774: 📝 Self-hosted backend docs show an Auth.js config that cannot work outside Next.js
### 💫 Describe the Issue With the Current Docs

The Netlify Functions and Vercel Functions backend pages both show a Tina backend config that cannot work as written. Following either page produces a 500 with:

```
(0 , import_credentials.default) is not a function
```

`tinacms-authjs` is published as ESM and default-imports `next-auth`, which is CommonJS. Webpack (and therefore Next.js) unwraps that via `__esModule`, but Node's own ESM loader and esbuild do not, so the imported value is the module object rather than the function. The Next.js backend page is unaffected; every other host is not.

Nothing in the docs mentions this, and changing `node_bundler` between `esbuild` and `zisi` does not help, so people lose hours before asking in Discord. Reported there on 15 Aug by a user on Astro + Netlify Functions.

### 📝 Outline the Proposed Change

1. On `/docs/reference/self-hosted/auth-provider/authjs/`, there is no guidance for using the package outside Next.js. Add a section covering it, with a working interop workaround.
2. On `/docs/reference/self-hosted/tina-backend/netlify-functions/`, the sample fails at module load. Add a callout quoting the exact error so it is searchable, linking to that section.
3. On `/docs/reference/self-hosted/tina-backend/vercel-functions/`, same as above.

### 💼 Tasks

- [ ] Add a "Using Auth.js outside Next.js" section to the Auth.js auth provider page
- [ ] Link to it from the Netlify Functions page
- [ ] Link to it from the Vercel Functions page
- [ ] Remove the section once the upstream fix ships

### ➕ More Information

The underlying package bug is tracked in tinacms/tinacms#7434. Once that is released, this docs section should be deleted rather than left to rot.

Two unrelated problems on the same pages, worth separate issues:

- The Vercel Functions sample uses `authentication:`, `AuthJsBackendAuthentication` and `LocalBackendAuthentication`. None of those exist any more; the current names are `authProvider:`, `AuthJsBackendAuthProvider` and `LocalBackendAuthProvider`.
- The Netlify sample's `'/api/tina/*'` route throws `PathError: Missing parameter name` on express 5, which is what `npm i express` installs today. It needs `'/api/tina/*splat'`, or the page should pin express 4.

kulesy:
Closes #4774

**TL;DR** Document the Auth.js interop workaround that self-hosted backends need outside Next.js.

**Pain:** anyone following the Netlify Functions or Vercel Functions backend docs gets a 500 with `(0 , import_credentials.default) is not a function`, because `tinacms-authjs` ships ESM that default-imports CommonJS `next-auth`. Webpack unwraps that automatically so the Next.js path works, but Node's own ESM loader and esbuild do not, which means both pages currently show a config that cannot work as written. There is no mention of this anywhere in the docs, and switching `node_bundler` does not help, so people burn hours before asking in Discord.

**Solution:** adds a "Using Auth.js outside Next.js" section to the Auth.js provider page with a workaround verified against `tinacms-authjs` 24.0.2 and `next-auth` 4.24.15, and links to it from the Netlify and Vercel pages using the exact error string so it is searchable. The upstream fix is tracked in tinacms/tinacms#7434, after which this section can be deleted.

### General Contributing:

### All New Content Submissions: (To be confirmed by reviewer)

- [x] Title is short & specific
- [x] Headers are logically ordered & consistent
- [x] Purpose of document is explained in the first paragraph
- [x] Procedures are tested and work
- [x] Any technical concepts are explained or linked to
- [x] Document follows structure from templates
- [x] All links work
- [x] The spelling and grammar checker has been run
- [x] Graphics and images are clear and useful
- [x] Any prerequisites and next steps are defined.

### DIFF
--- a/content/docs/reference/self-hosted/auth-provider/authjs.mdx
+++ b/content/docs/reference/self-hosted/auth-provider/authjs.mdx
@@ -96,6 +96,65 @@ export default (req, res) => {
 }
 ```
 
+## Using Auth.js outside Next.js
+
+<WarningCallout
+  body={<>
+    **Known issue:** `tinacms-authjs` currently only works when your backend is bundled by Next.js. On any other host (Netlify Functions, Vercel Functions, or plain Node), the backend fails with `(0 , import_credentials.default) is not a function`, followed by `NextAuth is not a function` once that is worked around. Tracking issue: [tinacms/tinacms#7434](https://github.com/tinacms/tinacms/issues/7434). Until it is fixed, apply the workaround below.
+  </>}
+/>
+
+`tinacms-authjs` is published as ESM and does a default import of `next-auth`, which is CommonJS. Webpack (and therefore Next.js) unwraps that automatically, but Node's own ESM loader and esbuild do not, so the imported value is the module object rather than the function.
+
+To work around it, build the credentials provider yourself and swap in the auth route handler, both using an interop-safe unwrap:
+
+```js
+import NextAuthImport from 'next-auth'
+import CredentialsImport from 'next-auth/providers/credentials'
+
+const NextAuth = NextAuthImport.default ?? NextAuthImport
+const Credentials = CredentialsImport.default ?? CredentialsImport
+
+const provider = Credentials({
+  credentials: {
+    username: { label: 'Username', type: 'text' },
+    password: { label: 'Password', type: 'password' },
+  },
+  authorize: async ({ username, password }) => {
+    const result = await databaseClient.authenticate({ username, password })
+    return result.data?.authenticate || null
+  },
+})
+// This name must match exactly, or sign-in will not route to this provider
+provider.name = 'TinaCredentials'
+
+const authOptions = TinaAuthJSOptions({
+  databaseClient,
+  secret: process.env.NEXTAUTH_SECRET,
+  providers: [provider],
+})
+
+const authJs = AuthJsBackendAuthProvider({ authOptions })
+authJs.extraRoutes.auth.handler = async (req, res, opts) => {
+  const url = new URL(req.url, `http://${req.headers?.host || 'localhost'}`)
+  req.query.nextauth = url.pathname
+    ?.replace(`${opts.basePath}auth/`, '')
+    ?.split('/')
+  await NextAuth(authOptions)(req, res)
+}
+```
+
+Then pass `authJs` to `TinaNodeBackend` in place of the inline `AuthJsBackendAuthProvider({ ... })` call, keeping your `isLocal` branch:
+
+```js
+const tinaBackend = TinaNodeBackend({
+  authProvider: isLocal ? LocalBackendAuthProvider() : authJs,
+  databaseClient,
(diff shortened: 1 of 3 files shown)


kulesy:
**Screenshots**

New section on the Auth.js auth provider page:

Known issue callout in the new Using Auth.js outside Next.js section

Figure: The "Known issue" callout that opens the new section

Pointer added to the Netlify Functions page, directly under the config sample:

Callout on the Netlify Functions page

Figure: The same warning on the Netlify Functions page, quoting the literal error string

The Vercel Functions page gets the same callout, minus the `node_bundler` sentence.

**How the workaround was verified**

Not just typed out. I rebuilt the docs' Netlify function (express 4 + cors + cookie-parser + `serverless-http` + `TinaNodeBackend` + `AuthJsBackendAuthProvider`), bundled it with `esbuild --bundle --platform=node --format=cjs` the way zip-it-and-ship-it does, and drove real API Gateway events through the handler with a cookie jar. Only `databaseClient` was stubbed, implementing the two operations `tinacms-authjs` actually calls.

Control, the docs config unchanged, dies at module load:

```
TypeError: (0 , import_credentials.default) is not a function
    at TinaCredentialsProvider
    at TinaAuthJSOptions
```

With the documented workaround:

| Request | Result |
| --- | --- |
| `GET /auth/session` logged out | `200 {}` |
| `GET /auth/csrf` | `200` + token |
| `POST /auth/callback/credentials` good creds | `200`, session cookie set |
| `GET /auth/session` after login | `200` with `role: "user"` |
| `POST /api/tina/gql` with cookie | `400 {"error":"no variables"}` |
| `POST /api/tina/gql` no cookie | `401 {"error":"Unauthorized"}` |
| `POST /auth/callback/credentials` wrong password | `401 CredentialsSignin` |

The `400` vs `401` pair is the important one: with a session cookie the request clears `isAuthorized` and reaches the GraphQL handler, which rejects my deliberately empty body; without it, `401`. So the patched provider still guards routes. The CommonJS variant of the two added imports was tested separately and behaves identically.

**Rendering checks**

MDX compiles on all three pages, the anchor `using-authjs-outside-nextjs` matches what `getDocId` generates from the heading, the ToC picks up the new heading, and both callout links resolve. The only console errors on these pages are pre-existing styled-components prop warnings in the ToC component.

**Notes for the reviewer**

- `docs-zh` is untouched; the three pages have Chinese counterparts that need translating separately.
- This section should be deleted once tinacms/tinacms#7434 ships, rather than left to rot.
- Two unrelated breakages on the same pages are written up in #4774 and deliberately left out of this PR: the Vercel sample uses API names that no longer exist (`AuthJsBackendAuthentication`, `LocalBackendAuthentication`, `authentication:`), and the Netlify sample's `'/api/tina/*'` route throws `PathError` on express 5.

PR #4795, opened 2026-08-19
docs: fix three pages that render as raw source

### LINKED ISSUE #4794: 📝 Three docs pages render as raw source because their MDX fails to parse
### Pain

Three docs pages fail to parse in `@tinacms/mdx`, so the renderer falls back to a single `invalid_markdown` node and prints the raw file body inside one `<pre>`. Readers get unparsed `##` headings, literal backticks, unrendered code fences and a raw `<WarningCallout ... />` tag instead of a page.

Affected:

- https://tina.io/docs/contextual-editing/astro
- https://tina.io/zh/docs/contextual-editing/astro
- https://tina.io/docs/errors/esbuild-error

Two causes:

1. The Astro pages reference `` `<TinaIsland>` `` inside a `WarningCallout` body. `body={<>...</>}` is a JSX attribute expression, so acorn parses its contents as real JSX before Tina reinterprets the children as markdown, and code span backticks give no protection. Acorn fails with `Expected corresponding JSX closing tag for <TinaIsland>` and the whole file collapses.
2. `esbuild-error.mdx` has a bare `{ts,js}` in a heading, which MDX reads as a JS expression: `Unexpected expression ts,js`.

### Approach

Self-close the tag as `` `<TinaIsland />` `` so acorn accepts it while the rendered text keeps its angle brackets. Backtick the heading to `` `config.{ts,js}` ``, matching how that string is already written in the page's opening paragraph.

### Acceptance Criteria

- [ ] All three pages render parsed markdown, with no page-wide `<pre>` block
- [ ] Running `parseMDX` over every file under `content/` reports zero `invalid_markdown` results (currently 3 of 809 fail)

kulesy:
Closes #4794

**TL;DR** Fix three docs pages that render as a wall of raw markdown inside a single `<pre>`.

**Pain:** When `@tinacms/mdx` cannot parse a document it emits one `invalid_markdown` node and the renderer prints the raw body verbatim, so readers get unparsed headings, literal backticks and a raw `<WarningCallout ... />` tag instead of a page. The Astro visual editing docs (en and zh) reference `` `<TinaIsland>` `` inside a `WarningCallout` body, where acorn parses the contents as real JSX and code span backticks give no protection, and `esbuild-error.mdx` carries a bare `{ts,js}` in a heading that MDX reads as a JS expression.

**Solution:** Self-close the tag as `` `<TinaIsland />` `` so acorn accepts it while the rendered text keeps its angle brackets, and backtick the heading to `` `config.{ts,js}` `` to match how that same string is already written in the page's opening paragraph.

### General Contributing:

### DIFF
--- a/content/docs-zh/contextual-editing/astro.mdx
+++ b/content/docs-zh/contextual-editing/astro.mdx
@@ -30,7 +30,7 @@ Astro 的可视化编辑路径不使用 `useTina()`；该钩子是 React 特有�
     **您需要一个 SSR 适配器。** 每个岛屿的刷新端点 (`/tina-island/[name]`)
     在每次按键时运行。将 `adapter` 设置在 `astro.config.mjs` 中为
     `@astrojs/node`、`@astrojs/vercel`、`@astrojs/netlify` 或 `@astrojs/cloudflare`。
-    `output: 'server'` 是最简单的选择；`output: 'static'` 也可以，只要您将可编辑区域包裹在 [`<TinaIsland>`](#static-site-editing) 中。
+    `output: 'server'` 是最简单的选择；`output: 'static'` 也可以，只要您将可编辑区域包裹在 [`<TinaIsland />`](#static-site-editing) 中。
   </>}
 />
 
--- a/content/docs/contextual-editing/astro.mdx
+++ b/content/docs/contextual-editing/astro.mdx
@@ -32,7 +32,7 @@ In production (no admin parent), the middleware injects nothing and `init()` exi
     runs at request time on every keystroke. Set `adapter` in `astro.config.mjs` to
     `@astrojs/node`, `@astrojs/vercel`, `@astrojs/netlify`, or `@astrojs/cloudflare`.
     `output: 'server'` is the simplest choice; `output: 'static'` also works as
-    long as you wrap editable regions in [`<TinaIsland>`](#static-site-editing).
+    long as you wrap editable regions in [`<TinaIsland />`](#static-site-editing).
   </>}
 />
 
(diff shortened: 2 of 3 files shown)


kulesy:
Verified with `parseMDX` from `@tinacms/mdx` (2.1.9) across all 809 files under `content/`: 3 failed to parse before this change, 0 after. The Astro page goes from a single `invalid_markdown` node to 64 parsed blocks.

Acorn errors captured before the fix:

- `content/docs/contextual-editing/astro.mdx` and `content/docs-zh/contextual-editing/astro.mdx`: `Expected corresponding JSX closing tag for <TinaIsland>`, surfacing at file level as `Unexpected token }`
- `content/docs/errors/esbuild-error.mdx`: `Unexpected expression ts,js`

Before shots are current production, after shots are the Vercel preview for e7d4539.

### Visual Editing Setup (Astro)



Figure: Before. The whole document is dumped as raw markdown inside one `<pre>`, overflowing horizontally, with the `<WarningCallout ... />` tag and the ` ```bash ` fences visible as literal text. The "On This Page" panel is empty because no headings parsed.



Figure: After. Prose, code spans, the ordered list and the SSR-adapter callout all render, and the "On This Page" nav is populated again.

### ESbuild compilation and build errors



Figure: Before. Same failure mode from the bare `{ts,js}` in the second heading.



Figure: After. Both `ERROR:` headings render, and the two `ts` code blocks pick up syntax highlighting.

Escaping to `&lt;TinaIsland&gt;` also parses, but the entity survives into the rich-text AST as literal text and renders as `&lt;TinaIsland&gt;`, which is why the self-closing form was used instead.

isaaclombardssw (approved):
😱 nice catch

PR #4806, opened 2026-08-24
update security files

joshbermanssw:
As per annual security email


RobJT-wq (approved):
✅ Confirmed that this is the correct Public Key for Tina.io

PR #4809, opened 2026-08-24
🐛 Fix the v3.12.1 release notes crashing the dev server

joshbermanssw:
## TL;DR

`pnpm dev` and `tinacms build` both die on `main` with `TypeError: str.split is not a function` during "Building search index". `content/whats-new-tinacms/v3.12.1.json` stores the whole changelog as one rich-text object in `changesDescription`, a field the schema declares as `string`; this splits it into the ten string entries the field expects.

## Why

The generated file has a single `changesList` entry whose description is a rich-text AST:

```json
"changesDescription": { "type": "root", "children": [ … ] }
```

Every other release file stores one plain string per change, with its own PR, commit and author metadata. The search indexer calls `.split()` on that value and throws before the server starts, so this blocks anyone running the site locally, and the next deploy from `main`.

Reproduced against the installed indexer:

```
as committed        : THREW -> str.split is not a function
with a string value : OK
```

## Reviewer notes

- **Content only.** Nothing changed in the schema or the site code.
- **Ten entries, no dedupe.** `#7469` and `#7445` appear more than once — those are separate changesets from the same PR, matching how the changelog reports them.
- **Full text preserved.** Continuation paragraphs are kept, minus the two-space indent the changeset format adds.
- **Cause not yet pinned.** `scripts/create-release-notes.sh` branches on whether the payload parses as JSON. Which branch produced this shape is still open — worth a look before the next release, since it will recur.

## Tests

Scanned every JSON collection for values whose type disagrees with the schema: this file was the only one, and there are none left. `tinacms build --local` now completes.

## Links

- Generated by [#4807](https://github.com/tinacms/tina.io/pull/4807)

### DIFF
--- a/content/whats-new-tinacms/v3.12.1.json
+++ b/content/whats-new-tinacms/v3.12.1.json
@@ -3,23 +3,97 @@
   "dateReleased": "2026-08-24T07:06:00Z",
   "changesObject": [
     {
-      "changesTitle": "Release Notes",
+      "changesTitle": "Patch Changes",
       "changesList": [
         {
-          "changesDescription": {
-            "type": "root",
-            "children": [
-              {
-                "type": "p",
-                "children": [
-                  {
-                    "type": "text",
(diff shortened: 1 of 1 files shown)


suiyangqiu (approved):
LGTM

PR #4824, opened 2026-08-28
📝 Add a blog post on semantic table rendering in TinaMarkdown

joshbermanssw:
## TL;DR

`tinacms` is shipping semantic `<thead>`/`<th>` for markdown tables ([tinacms#7169](https://github.com/tinacms/tinacms/issues/7169)), which changes the rendered HTML on every existing site. This post announces it, explains why it ships as a minor rather than a major, and shows the `th` component override people will need to add.

## Why

The change is breaking for anyone styling table headers, but the public API is untouched, so it goes out as a minor. That decision needs to be said out loud somewhere public before the release lands.

## Links

- [Issue: Render semantic thead/th for markdown tables](https://github.com/tinacms/tinacms/issues/7169)

---

Drafted with Claude Opus 5 in Claude Code.


isaaclombardssw (dismissed):
This is solid, only comment is we don't really need the related links at the bottom.

isaaclombardssw (approved):
✅

isaaclombardssw (commented):
Approved, but two things to sort before this goes live.

1. tinacms/tinacms#7494 is still open, and the latest release is `tinacms@3.12.1`. The post says "One we merged recently" and tells people to check their tables "before you deploy TinaCMS 3.13" - neither is true yet. Hold the publish until it's merged and 3.13 is out.

2. The changeset has a second breaking change the post doesn't mention: the default inline `border: 1px solid #EDECF3` on `<table>` and the `border`/`padding: 0.25rem` on cells are gone, and `align` is no longer leaked to the DOM. That's actually what the before/after screenshots show - the boxed grid disappearing is the inline styles going away, not the header semantics. As written the post attributes the visual change to `thead`/`th`, so anyone who was relying on those inline defaults gets no warning and no fix.

Rest checks out - @lastenvoy30's first and only PR to the repo, the ~18 outside contributors over 12 months, and the rich-text editor already emitting `thead`/`th`.

wicksipedia (approved):

PR #4841, opened 2026-09-07
Stop reading every blog post body to build the blog lists

wicksipedia:
Every build logs a run of these warnings:

```
Failed to set fetch cache https://content.tinajs.io/... Error: Failed to set
Next.js data cache, items over 2MB can not be cached (15841626 bytes)
```

Two blog helpers ask Tina for the whole post collection, bodies and all, and then use a small part of the answer. Next.js refuses to cache a response over 2 MB, so it logs and moves on. The build still passes, so this is noise rather than a failure, but it is a lot of data to move on every build.

## What changed

`generateBlogStaticParams` reads only the filename of each post. It now uses a query that returns filenames and page info.

`getBlogIndexPosts` sorts the whole collection but shows eight posts per page. Only those eight need a body, for the 200 character excerpt. It now reads the list without bodies, then reads the body of each post on the page.

Both use new queries in `tina/queries/post.gql`.

## Measurements

Response size for the whole post collection, taken against a local Tina server:

| Helper | Before | After |
| --- | --- | --- |
| `generateBlogStaticParams` | 3,894,068 bytes | 3,300 bytes |
| `getBlogIndexPosts` | 3,894,068 bytes | about 105,000 bytes |

The second figure is one list query of 29,477 bytes plus one body per shown post, at about 9,400 bytes each.

TinaCloud responses run three to four times larger than local ones, which matches the 15 MB in the build log.

## Testing

- `npx tsc --noEmit` passes
- `pnpm test` passes, 82 tests across 15 suites
- `npx biome check` is clean
- `/blog`, `/blog/page/2`, `/zh/blog` and a single post page all return 200

## Two things to check on this run

1. Whether the warning count drops. I could only measure against a local server, so I cannot prove the TinaCloud figures from here.
2. Whether build time moves. Recent runs on other branches took 14m29s to 15m24s. `getBlogIndexPosts` now makes eight small requests per index page instead of one large one, so the build could get slower even though it moves far less data.

## Known, not caused by this change

The excerpt block on `/blog` renders empty. It is also empty on `main` before this change. Worth a separate look.

https://claude.ai/code/session_01EQtgguUUPWKQqAE8DZt8JU


joshbermanssw (approved):
LGTM

PR #4867, opened 2026-09-09
📝 Document the rich-text parser, skipEscaping and isBody options

kulesy:
No linked issue - the missing options came up in https://github.com/tinacms/tinacms/discussions/7527

**TL;DR** Adds `parser`, its `skipEscaping` option and `isBody` to the rich-text field reference, and says plainly that the parser default comes from the collection's `format` rather than the field. Nothing existing changes and the `overrides` group is untouched.

**Pain:** `parser` is the fix for a whole class of "Tina mangled my markdown on save" reports, like the discussion above where `$a + b_{c}$` is written back as `$a + b\_{c}$`, but it appears nowhere in the docs. `isBody` is used four times in the examples further down the same page and never defined.

**Solution:** documents all three in the existing `apiReference` block, with a note that the default is set by the collection format. The Markdown Syntax Guide also gets a one-line "Escaped markdown entities" entry under Automatic Transforms, pointing back at the option.

`content/docs-zh/reference/types/rich-text.mdx` mirrors this page and is deliberately left alone here, for the auto-translation workflow to pick up.

### General Contributing:

### DIFF
--- a/content/docs/reference/markdown-spec.mdx
+++ b/content/docs/reference/markdown-spec.mdx
@@ -209,4 +209,8 @@ From `- Item 1` to `Item 1`.
 
 From `* > My blockquote` to `* My blockquote`.
 
+### Escaped markdown entities
+
+From `$a + b_{c}$` to `$a + b\_{c}$`. Turn this off with `parser: { type: 'markdown', skipEscaping: 'all' }` on the [rich-text field](/docs/reference/types/rich-text).
+
 ###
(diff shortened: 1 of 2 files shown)


kulesy:
Before and after of the Type Definition table, both from the dev server at the same width, with the parser group expanded in the after shot so the new rows show.

the rich-text Type Definition table on main

Figure: before, on main. type, name, templates and the overrides group, live at https://tina.io/docs/reference/types/rich-text

the rich-text Type Definition table with the new rows

Figure: after, with isBody, the parser group and the two notes under the table, live at https://tina-8pt8ekhkg-tinacms.vercel.app/docs/reference/types/rich-text

The other change is one line in the Markdown Syntax Guide: https://tina-8pt8ekhkg-tinacms.vercel.app/docs/reference/markdown-spec

isaaclombardssw (approved):
✅

PR #4870, opened 2026-09-09
fix projects in pricing (PR from TinaCMS)

0xharkirat:
Projects column on pricing page shows 1 project on all plans,

As per my conversation with @kulesy , we can have unlimited projects in any plan, so updated the text


kulesy (approved):

wicksipedia (changes requested):
no, this change will suggest that you get unlimited projects on the plan

PR #4876, opened 2026-09-11
✨ Proxy /docs to the new docs instance behind DOCS_UPSTREAM_URL

joshbermanssw:
cc: —
Related PBI: https://github.com/tinacms/tinacms/issues/7552

Stack: this PR → #4877 (search/sitemap/workflow prep) → deletion PR (after cutover).

## TL;DR

tina.io/docs is moving to a separate tina-docs instance, but nothing on tina.io routes there yet. This PR adds an env-gated `beforeFiles` rewrite so `/docs/*` and `/zh/docs/*` proxy to `DOCS_UPSTREAM_URL` when it is set, and changes nothing when it is unset.

## Verify

On the Vercel preview for this branch, set `DOCS_UPSTREAM_URL=https://docs-three-lovat.vercel.app` and load `/docs`, `/docs/admin`, and `/zh/docs`. Each should proxy to the docs instance. Redirects still run before these rewrites (Next checks headers, then redirects, then `beforeFiles` rewrites), and `middleware.ts` (matcher `/`) is untouched by `/docs/*` requests.

## Links

- [Epic: migrate docs from tina.io](https://github.com/tinacms/tinacms/issues/7552)

---
Changes made by Claude Sonnet 5 via Claude Code.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4877, opened 2026-09-11
🔍 Stop indexing and listing docs ahead of tina-docs cutover

joshbermanssw:
cc: —
Related PBI: https://github.com/tinacms/tinacms/issues/7552

Stacked on #4876 (`jb/docs-upstream-rewrite`). Merge order at cutover: set `DOCS_UPSTREAM_URL` in production, merge #4876, then merge this PR in the same window; merging this alone would leave tina.io's still-local docs pages without docs search.

## TL;DR

Once the tina-docs instance takes over `tina.io/docs`, tina.io must stop treating `content/docs` as its own: it still indexes docs into Algolia, lists them in its sitemap, auto-translates them on merge, and shows docs result tabs in the search UI. This PR removes all four and points robots at the new instance's own sitemap.

## Verify

This PR merges together with tinacms/tina.io#4876 at cutover; on its own it's inert since docs still live under `app/docs` today. To check locally: `pnpm type-check` and `pnpm lint` are clean, `/search` and the header's quick-search overlay both render blog results only (no empty "DOCS" tab), and `indices/createIndices.ts` no longer reads `content/docs`. Note: `app/docs` still imports `SearchNavigation.tsx` (the overlay component) and keeps working unchanged until the post-cutover deletion PR removes `app/docs` entirely.

## Links

- [Epic: migrate docs from tina.io](https://github.com/tinacms/tinacms/issues/7552)

---
Changes made by Claude Sonnet 5 via Claude Code.


components/Docs/docsSearch/SearchNavigation.tsx
```diff
@@ -242,7 +194,7 @@ export const SearchModal = ({
               ref={inputRef}
               type="text"
               className="w-full py-3 px-6 rounded-full border border-gray-300 bg-white shadow-md focus:outline-none focus:ring-2 focus:ring-orange-500 focus:border-transparent"
-              placeholder="Search docs and blogs..."
+              placeholder="Search blog posts..."
```
kulesy, line 197: The moment this merges the search box on every docs page stops finding docs, because /docs is still served from here until someone sets DOCS_UPSTREAM_URL and redeploys.

next-sitemap.config.js
```diff
@@ -12,6 +12,7 @@ module.exports = {
     additionalSitemaps: [
       'https://tina.io/tinadocs/sitemap.xml',
       'https://tina.io/tinadocs/doc/sitemap.xml',
+      'https://tina.io/docs/sitemap.xml',
```
kulesy, line 15: This sitemap 404s today and also 404s on the docs instance it will proxy to, so with the excludes below the docs end up listed in no sitemap at all.

components/Docs/docsSearch/SearchNavigation.tsx
joshbermanssw: Agreed, and that is why this PR is stacked on #4876 and marked to merge together with it at cutover, with `DOCS_UPSTREAM_URL` set in production first. Once the rewrite is live tina.io no longer renders any docs page (this component included), so the window where docs search would be dead on locally served pages does not exist as long as the merge order holds. I have added that to the PR body explicitly.

next-sitemap.config.js
joshbermanssw: Good catch. The instance was emitting its sitemap at `/docs/doc/sitemap.xml` (template quirk) with relative `<loc>` entries because `NEXT_PUBLIC_SITE_URL` was unset on Vercel. Fixed on the instance in https://github.com/tinacms/docs/pull/34: it now serves `https://tina.io/docs/sitemap.xml` with absolute URLs and without the `/r/` alias redirects, so the `additionalSitemaps` entry here resolves once the rewrite is live.

kulesy (approved):

PR #4879, opened 2026-09-11
🧹 Remove the docs routes, schema, content and media from tina.io

joshbermanssw:
cc: —
Related PBI: https://github.com/tinacms/tinacms/issues/7552

Stacked on #4877; merge only after the rewrite has been live in production for at least 48 hours.

## TL;DR

Once the `DOCS_UPSTREAM_URL` rewrite from #4876 serves `tina.io/docs` and `tina.io/zh/docs` from the tina-docs instance, everything docs-related in tina.io is dead weight that still builds, indexes and ships. This PR deletes it: the `app/docs` and `app/zh/docs` routes, `components/Docs`, `utils/docs`, the `doc`, `docZh` and `docsTableOfContents` collections (with a regenerated `tina-lock.json`), 424 content files, 222 media files (48.9 MB) and the 97 redirects in `content/settings/config.json` whose source starts with `/docs`.

- `/search` keeps working as a blog-only search; the modal and results components moved to `components/search/`, and the page no longer renders the docs sidebar around the results.
- Media was deleted by script: every tracked file under `public/` referenced by the docs content, or under `public/img/docs`, was checked against the rest of the repo. Four `public/img/docs` files stay because blog posts and `tina/collectionsSchema/examples.tsx` still reference them.
- `scripts/cms-reference-check` walked `content/docs`, so it goes too; port it to tinacms/docs if it is still wanted.

## Links

- [Epic: migrate docs from tina.io](https://github.com/tinacms/tinacms/issues/7552)
- [#4876: rewrite /docs to the instance](https://github.com/tinacms/tina.io/pull/4876)
- [#4877: cutover prep](https://github.com/tinacms/tina.io/pull/4877)

---
Changes made by Claude Opus 5 via Claude Code.


content/settings/config.json
```diff
@@ -113,211 +72,21 @@
...
-      "destination": "/guides/nextjs/git/getting-started",
-      "permanent": true
-    },
-    {
-      "source": "/docs/nextjs/adding-backends",
-      "destination": "/docs/reference/self-hosted/tina-backend/nextjs",
-      "permanent": true
-    },
-    {
-      "source": "/docs/nextjs/creating-forms",
```
kulesy, line 267: Dropping these two takes a pair of URLs that work today down to a 404, because the docs instance serves nothing at either path.

joshbermanssw: Traced it end to end. After cutover the instance owns these two: `GET /docs/nextjs/bootstrapping` on the instance already answers 308 to `/guides/nextjs/git/getting-started` (its redirect is `basePath: false`, so the Location is the tina.io path), tina.io's kept `/guides/nextjs(.*)` rule then sends that to `/docs/guides/`, and `/docs/guides` is 200 on both hosts today. Same for `creating-forms` → `/guides/nextjs/git/creating-git-forms`. So removing them here does not create a 404; the instance copy in `tinacms/docs` `content/settings/config.json` is the live one. Happy to keep the pair on tina.io as belt and braces if you would rather, but they would be dead once the rewrite is in front.

kulesy (approved):

PR #4888, opened 2026-09-14
Make the plan comparison table follow the billing toggle

kulesy:
**TL;DR** Moves the plan comparison table inside the pricing block as a field, so the monthly/annual toggle is plain component state shared by the cards and the table, and gives table columns an optional annual by-line. The table header now switches between the monthly and annual price with the toggle; the standalone table block still exists for other pages and renders exactly as before.

**Pain:** on tina.io/pricing, switching the toggle to Annually drops the plan cards to $24, $41 and $249 while the comparison table header keeps saying $29, $49 and $299 per project per month, so the two halves of the page disagree.

**Solution:** adds a comparison table field to the pricing template, moves the pricing page's table content into it, renders it under the cards with the toggle state passed in, and adds an annual by-line per column. The page order changes slightly: the table now sits directly under the cards, ahead of the customer logos.

### General Contributing:


kulesy:
Table header from a local build at 1440 wide, in the three states that matter, plus where the table now sits on the page. Images are on a non-merging screenshots branch.

Comparison table header before the fix and after, on Annually and on Monthly

Figure: before, the header stays at the monthly price whatever the toggle says; after, it follows the toggle, live at https://tina-khnbuv2tq-tinacms.vercel.app/pricing

Pricing cards with the comparison table directly beneath them

Figure: the table now renders inside the pricing block, under the cards and ahead of the customer logos

joshbermanssw (approved):
LGTM

PR #4894, opened 2026-09-20
Serve the Chinese site from tinaio.cn without a /zh prefix

ZenoWang1999:
Moves the Chinese site onto its own domain. `tinaio.cn` and `tina.io` serve the same path shape — `/pricing` is the Chinese page on one and the English page on the other — so `/zh` no longer appears in a public URL.

Middleware decides locale from the hostname and rewrites the Chinese domain onto the existing `/zh/...` routes, leaving the `app/` tree unchanged. Any other host returns on the first line with no rewrite, redirect or cookie, so English is untouched.

Also included: the 21 missing Chinese pages, `/zh` dropped from 667 links in Chinese content, and the ICP number updated for tinaio.cn.

### Behaviour changes

- The home page no longer redirects based on cookie or `Accept-Language`.
- Language switching is a cross-origin navigation that keeps the current page.
- `/zh` stays valid and is only collapsed in production, so Chinese pages remain openable locally.
- not-found takes its locale from the request host instead of the URL prefix.

### Verified

Both Chinese hosts serve the home, pricing and docs pages; `tina.io` is unchanged and `tina.io/zh` still resolves. English content files are untouched and all 1033 external URLs in Chinese content are byte-identical. `tsc`, `biome` and `jest` pass.

### Before merging

Add `tinaio.cn` and `www.tinaio.cn` as Vercel custom domains, and make sure the reverse proxy forwards the browser-facing hostname — if it sends the upstream origin, the Chinese site renders in English.

hreflang and per-domain sitemaps are a follow-up: `next-sitemap.config.js` still hard-codes `https://tina.io`.

### General Contributing:

https://claude.ai/code/session_01PQhrmU6QDCgHXmgPDsKBQR


tino-liu (changes requested):
Image
**Figure: Weird / in the home page**

tino-liu (approved):
kk lgtm

